PR #381, opened 2014-02-28
Addind cleanstate funtions to euca

obino:
Euca didn't have the cleanup function implemented, which means that groups
and keys were left behind.


obino:
Not needed anymore since we switched to use the ec2_agent, and it correctly deletes the groups and keyname.

PR #401, opened 2014-07-03
Make sure the 'cert' directory exists.

obino:
While working on a docker instance, I got into an error copying the certs over since the cert directory didn't exist on the target. This was tested on that docker instance.


obino:
We do have a docker image already. CLosing this pull.

PR #423, opened 2015-02-11
Bump version to 2.2.0

nlake44:


nlake44:
Accidentally called it version 2.1.0, but should be 2.2.0.

PR #431, opened 2015-04-08
Don't override an init.d script.

obino:
The init.d script will be put in place at install time, no need to
override it. The script will not start appscale-controller unless the
secret is been put in place (ie AppScale is configured).


nlake44:
:+1:

PR #435, opened 2015-04-13
Fix inaccurate abort messages

cdonati:
Previously, it would always say "AppScale termination was cancelled" even if the user was aborting an "appscale up."


cdonati:
I tested with EC2 and GCE blacklisted instance types.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/452/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/859/

PR #442, opened 2015-04-21
Ssh keys

obino:
Don't override authorized_keys and remove duplicates.

### DIFF
--- a/lib/remote_helper.py
+++ b/lib/remote_helper.py
@@ -321,6 +321,8 @@ class RemoteHelper(object):
         cls.ssh(host, keyname, "sudo sed -n '/.*Please login/d; " \
           "w/root/.ssh/authorized_keys' {0}".format(temp_file.name),
           is_verbose, user=getpass.getuser())
+        cls.ssh(host, keyname, "sudo rm -f {0}".format(temp_file.name),
+          is_verbose, user=getpass.getuser())
         temp_file.close()
         return
       else:
@@ -336,6 +338,8 @@ class RemoteHelper(object):
         is_verbose, user='ubuntu')
       cls.ssh(host, keyname, "sudo sed -n '/.*Please login/d; " \
         "w/root/.ssh/authorized_keys' {0}".format(temp_file.name),
+        is_verbose, user='ubuntu')
+      cls.ssh(host, keyname, "sudo rm -f {0}".format(temp_file.name),
         is_verbose, user='ubuntu')
       temp_file.close()
     else:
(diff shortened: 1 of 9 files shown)


lib/remote_helper.py
```diff
@@ -313,8 +314,14 @@ def enable_root_login(cls, host, keyname, infrastructure, is_verbose):
       # logged-in user, so log in as that user to enable root login.
       if infrastructure == "gce":
         AppScaleLogger.log("Root login not enabled - enabling it now.")
-        cls.ssh(host, keyname, 'sudo cp ~/.ssh/authorized_keys /root/.ssh/',
+        temp_file = tempfile.NamedTemporaryFile()
```
cdonati: You probably don't need `tempfile` here since you don't need to make a temporary file on the machine running the tools. Maybe use `mktmp` as part of the ssh command.

lib/remote_helper.py
```diff
@@ -313,8 +314,14 @@ def enable_root_login(cls, host, keyname, infrastructure, is_verbose):
...
+        cls.ssh(host, keyname, "sudo sort -u ~/.ssh/authorized_keys " \
+          "/root/.ssh/authorized_keys -o {0}".format(temp_file.name),
           is_verbose, user=getpass.getuser())
+        cls.ssh(host, keyname, "sudo sed -n '/.*Please login/d; " \
+          "w/root/.ssh/authorized_keys' {0}".format(temp_file.name),
+          is_verbose, user=getpass.getuser())
```
cdonati: Would it make sense to make a single ssh call with `&&` separating the commands?

lib/remote_helper.py
```diff
@@ -313,8 +314,14 @@ def enable_root_login(cls, host, keyname, infrastructure, is_verbose):
       # logged-in user, so log in as that user to enable root login.
       if infrastructure == "gce":
         AppScaleLogger.log("Root login not enabled - enabling it now.")
-        cls.ssh(host, keyname, 'sudo cp ~/.ssh/authorized_keys /root/.ssh/',
+        temp_file = tempfile.NamedTemporaryFile()
+        cls.ssh(host, keyname, "sudo sort -u ~/.ssh/authorized_keys " \
```
cdonati: Don't the tools log in as root? Is `sudo` necessary?

lib/remote_helper.py
obino: Not sure how to make it easy to run mktemp remotely and catch the name of the file: pipes and redirections are tricky when using sudo. The remote file was left behind: I will have a patch to remove it. Thanks for catching it!

lib/remote_helper.py
obino: Personally I find long lines to be difficult to read and to debug. I will switch to IRC to check with you guys and see what's best.

PR #446, opened 2015-05-08
better error message when you put the wrong IP, and make the template…

nlake44:
… for cluster match the IP from VirtualBox


obino:
https://ocd.appscale.com:8080/job/Daily%20Build/124/ all good!

PR #465, opened 2015-11-12
Fix for all_ssh_key.

obino:


obino:
OCD run to check no regression somewhere else.

PR #466, opened 2015-11-18
Print the kind of exception we caught.

obino:
We now at least print the kind of exception we caught. Also make sure we
store the appcontroller crashlog for later viewing.

### DIFF
--- a/lib/appscale_logger.py
+++ b/lib/appscale_logger.py
@@ -100,7 +100,8 @@ class AppScaleLogger():
       conn.request('POST', '/upload', payload, cls.HEADERS)
       _ = conn.getresponse()
       conn.close()
-    except Exception:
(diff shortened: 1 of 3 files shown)


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/733/

this run is not against master, so we can collect some crashlog for sure to see if it works.

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/742/

menivaitsi:
Is there an example in the build log?

obino:
@menivaitsi  yes, if you look at https://ocd.appscale.com:8080/job/Hawkeye/2667/console and search for 

Saw Exception while looking for UAServer

you will see that's there (ie we capture the exception message).

PR #468, opened 2015-11-25
Don't look for the User App Server any longer

obino:
With the latest appscale the user app server sits on the login node. There is no need to discover it.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/784/

PR #472, opened 2015-12-05
Use system pyyaml

cdonati:
This pull removes the build scripts for Lucid and Wheezy since we won't be building system packages for them. As explained in the README, those systems can still install from source with `python setup.py install`.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/822/

PR #477, opened 2016-01-30
Don't use monit quit.

obino:
It seems to cause issues with some version of monit, creting a zombie copy
of monit to linger in the system.

### DIFF
--- a/lib/remote_helper.py
+++ b/lib/remote_helper.py
@@ -694,14 +694,9 @@ class RemoteHelper(object):
     cls.scp(host, keyname, cls.MONIT_APPCONTROLLER_CONFIG_FILE,
       '/etc/monit/conf.d/appscale-controller-17443.cfg', is_verbose)
 
-    # Start up monit.
-    cls.ssh(host, keyname, 'monit quit; ', is_verbose)
+    # Re-start up monit to make sure it is running and it did reloaded its
+    # configuration (and consequently starts the controller).
     cls.ssh(host, keyname, 'service monit start', is_verbose)
-    time.sleep(1)
-
-    # Start the AppController.
-    cls.ssh(host, keyname, 'monit start -g controller', is_verbose)
-    time.sleep(1)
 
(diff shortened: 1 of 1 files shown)


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/998/

lib/remote_helper.py
```diff
@@ -694,14 +694,9 @@ def start_remote_appcontroller(cls, host, keyname, is_verbose):
...
 
-    # Start up monit.
-    cls.ssh(host, keyname, 'monit quit; ', is_verbose)
+    # Re-start up monit to make sure it is running and it did reloaded its
+    # configuration (and consequently starts the controller).
     cls.ssh(host, keyname, 'service monit start', is_verbose)
-    time.sleep(1)
-
-    # Start the AppController.
-    cls.ssh(host, keyname, 'monit start -g controller', is_verbose)
```
menivaitsi: Is the controller already up at this point?

lib/remote_helper.py
```diff
@@ -694,14 +694,9 @@ def start_remote_appcontroller(cls, host, keyname, is_verbose):
     cls.scp(host, keyname, cls.MONIT_APPCONTROLLER_CONFIG_FILE,
       '/etc/monit/conf.d/appscale-controller-17443.cfg', is_verbose)
 
-    # Start up monit.
-    cls.ssh(host, keyname, 'monit quit; ', is_verbose)
+    # Re-start up monit to make sure it is running and it did reloaded its
```
menivaitsi: Change this to:

Re-start monit to make sure it is running and it reloads its

lib/remote_helper.py
obino: It shouldn't be running at this point, that's why we copied over the monit configuration. The restart of monit will start the controller.

obino:
It's not doing what is supposed to do: we shouldn't use monit at all to start the controller. Closing.

PR #486, opened 2016-02-09
Use only the terminate command on appscale down

cdonati:


cdonati:
This is meant to be merged with https://github.com/AppScale/appscale/pull/1933

PR #502, opened 2016-04-21
Ensure length of disk name does not exceed GCE max

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1360/

PR #505, opened 2016-05-03
Revert timeout

menivaitsi:
…nt side."

This reverts commit f2d2a98922f623cb340dba5b7c31b5bd62e6ffe8.

https://ocd.appscale.com:8080/job/Daily%20Build/1407/

### DIFF
--- a/lib/appscale_tools.py
+++ b/lib/appscale_tools.py
@@ -638,6 +638,8 @@ class AppScaleTools(object):
         break
       except ValueError:
         pass
+      except KeyError:
+        time.sleep(20)
     if not current_app:
       raise AppScaleException("Unable to get the serving port for the application.")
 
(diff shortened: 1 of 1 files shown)


lib/appscale_tools.py
```diff
@@ -636,14 +635,13 @@ def upload_app(cls, options):
...
+          continue
         break
       except ValueError:
         pass
       except KeyError:
-        pass
-      AppScaleLogger.verbose("Waiting {0} second(s) for a port to be assigned to {1}".\
-        format(sleep_time, app_id), options.verbose)
-      time.sleep(sleep_time)
+        time.sleep(20)
```
obino: Question: isn't 20 seconds a bit too long to wait for? Is this a path we expect to hit often?

menivaitsi: It's not a path we expect to hit often and because it really depends on the AC cycle and getting info in the app_info_map, it makes sense to me to be 20.

Happy to reconsider if you think otherwise.

PR #507, opened 2016-05-24
Adding upgrade command to the AppScale Tools

tmarballi:


bin/appscale-upgrade.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+# Programmer: Chris Bunch (chris@appscale.com)
```
menivaitsi: Remove this line.

lib/appscale.py
```diff
@@ -100,6 +107,8 @@ class AppScale():
   undeploy <appid>                  Removes <appid> from the current
                                     deployment. DATA ASSOCIATED WITH
                                     THE APPLICATION WILL BE LOST.
+  upgrade                           Upgrades AppScale code to its latest version
```
menivaitsi: I'd change the description to: 

> Upgrades AppScale to its latest version.

The data upgrade will probably not be a regular op and is part of the whole process.

lib/appscale.py
```diff
@@ -365,6 +374,29 @@ def get_all_ips(self, ips_layout):
 
     return all_ips
 
+  def get_ips_for_role(self, ips_layout, role):
+    """ Searched through the given IPs layout and finds all the unique
```
menivaitsi: Searched -> Searches

lib/appscale_tools.py
```diff
@@ -66,6 +69,22 @@ class AppScaleTools(object):
...
 
+  # Command to cd into the AppScale repository.
+  CD_APPSCALE_REPO = "cd ~/appscale"
+
+
+  # Command to run the Bootstrap.
+  RUN_BOOTSTRAP_COMMAND = "bash bootstrap.sh"
+
+
+  # String of commands to cd into the ApppScale repository and run the Bootstrap
```
menivaitsi: ApppScale -> AppScale

menivaitsi: Also don't forget the period at the end.

lib/appscale_tools.py
```diff
@@ -66,6 +69,22 @@ class AppScaleTools(object):
...
+  BOOTSTRAP_COMMAND = CD_APPSCALE_REPO + ";" + RUN_BOOTSTRAP_COMMAND
+
+
+  # Command to run the upgrade script from /appscale/scripts directory.
+  UPGRADE_SCRIPT = "python ~/appscale/scripts/datastore_upgrade.py"
```
menivaitsi: Instead of "~/appscale" can we use AppScale repo dir as a constant?
Create a new constant if there isn't one already.

That will help in making AppScale relocatable.

lib/appscale_tools.py
```diff
@@ -653,3 +672,48 @@ def upload_app(cls, options):
...
+      if response not in ['y', 'yes', 'Y', 'YES']:
+        raise AppScaleException("Cancelled AppScale upgrade.")
+      else:
+        AppScaleLogger.log("Shutting down AppScale...")
+        cls.terminate_instances(options)
+
+    AppScaleLogger.log("Upgrading AppScale to the latest version on "
+      "these machines: {}".format(options.ips))
+    AppScaleLogger.warn(("Running bootstrap on the machines to fetch latest " + \
+      "code and build AppScale. This will atleast take a few minutes."))
```
menivaitsi: "atleast take" -> "take at least"

lib/appscale_tools.py
```diff
@@ -653,3 +672,48 @@ def upload_app(cls, options):
...
+        AppScaleLogger.success("Successfully upgraded AppScale to its latest version.")
+      except ShellException:
+        AppScaleLogger.warn("Error executing bootstrap command to upgrade AppScale.")
+
+    zookeeper_ips = ""
+    for zk_ip in options.zk_ips:
+      zookeeper_ips += zk_ip + " "
+
+    upgrade_script_zk_loc = cls.UPGRADE_SCRIPT + " " + zookeeper_ips
+    AppScaleLogger.log("Upgrade script {}".format(upgrade_script_zk_loc))
```
menivaitsi: Since this prints out the command itself, should it be a debug message?
If you want you can leave the log and only print the zookeeper_ips that are being upgraded.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/1505/

appscale/tools/appscale_tools.py
```diff
@@ -62,11 +68,39 @@ class AppScaleTools(object):
...
+  # Command to run the Bootstrap.
+  RUN_BOOTSTRAP_COMMAND = "bash bootstrap.sh"
+
+
+  # String of commands to cd into the AppScale repository and run the Bootstrap.
+  BOOTSTRAP_COMMAND = "cd " + APPSCALE_REPO + ";" + RUN_BOOTSTRAP_COMMAND
+
+
+  # Command to run the upgrade script from /appscale/scripts directory.
+  UPGRADE_SCRIPT = "python " + APPSCALE_REPO + "/scripts/datastore_upgrade.py"
```
cdonati: For this to be a generic upgrade command, you probably don't want to point directly to a `datastore_upgrade.py` file. The server side should probably have a generic upgrade script that checks if an upgrade is even needed, what scripts to run, etc.

appscale/tools/appscale_tools.py
```diff
@@ -62,11 +68,39 @@ class AppScaleTools(object):
...
+
+
+  # Data Upgrade Success keyword from log file.
+  DATA_UPGRADE_SUCCESS = "Data upgrade status: SUCCESS"
```
cdonati: Same here. I wouldn't check for a specific string related to a data upgrade since later upgrades won't necessarily be data-related. One option is to write a response on the server side in JSON format that includes the success status and a more specific message if needed.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +688,62 @@ def upload_app(cls, options):
...
+      options: A Namespace that has fields for each parameter that can be
+        passed in via the command-line interface.
+    Returns:
+      A tuple containing the host and port where the application is serving
+        traffic from.
+    """
+    if os.path.exists(LocalState.get_secret_key_location(options.keyname)):
+      response = raw_input("AppScale needs to be down for this upgrade."
+        " Are you sure you want to proceed? (Y/N) ")
+      if response not in ['y', 'yes', 'Y', 'YES']:
```
cdonati: Just a tip: You can do `response.lower() not in ['y', 'yes']` here.

bin/appscale-upgrade.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
```
cdonati: For consistency with the other scripts, remove the `.py` extension.

bin/appscale-upgrade.py
```diff
@@ -0,0 +1,38 @@
...
+# Make sure we're on Python 2.6 or greater before importing any code
+# that's incompatible with older versions.
+import version_helper
+version_helper.ensure_valid_python_is_used()
+
+
+from appscale_logger import AppScaleLogger
+from appscale_tools import AppScaleTools
+from local_state import LocalState
+from parse_args import ParseArgs
```
cdonati: With the namespace package changes merged, you can do
```
from appscale.tools import version_helper
from appscale.tools.appscale_tools import AppScaleTools
```
etc.

You don't need the above try, except.

bin/appscale-upgrade.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+if __name__ == "__main__":
+  options = ParseArgs(sys.argv[1:], "appscale-upgrade").args
+  try:
+    AppScaleTools.upgrade(options)
+    sys.exit(0)
+  except Exception, e:
```
cdonati: Use the `as` syntax to differentiate between multiple exceptions and defining the exception variable.

appscale/tools/appscale.py
```diff
@@ -834,3 +865,35 @@ def register(self, deployment_id):
     AppScaleLogger.success(
       'Registration complete for AppScale deployment {0}.'
       .format(deployment['name']))
+
+  def upgrade(self):
+    """ Allows users to upgrade to the latest version of AppScale as well as
+    upgrades the data within zookeeper and cassandra.
+
+    Raises:
+      AppScaleException:
```
menivaitsi: Please remove Raises section here.

bin/appscale
```diff
@@ -176,6 +176,12 @@ elif command == "register":
 elif command in ["--version", "-v"]:
   print APPSCALE_VERSION
   sys.exit(0)
+elif command == "upgrade":
+  try:
+    appscale.upgrade()
+  except Exception, e:
```
menivaitsi: Please replace variable _e_ with something like _exception_ throughout this file.

appscale/tools/appscale.py
```diff
@@ -368,6 +376,29 @@ def get_all_ips(self, ips_layout):
 
     return all_ips
 
+  def get_ips_for_role(self, ips_layout, role):
```
cdonati: Is this function used anywhere?

appscale/tools/appscale.py
```diff
@@ -837,3 +868,29 @@ def register(self, deployment_id):
     AppScaleLogger.success(
       'Registration complete for AppScale deployment {0}.'
       .format(deployment['name']))
+
+  def upgrade(self):
+    """ Allows users to upgrade to the latest version of AppScale as well as
+    upgrades the data within zookeeper and cassandra. """
```
cdonati: This command will not always upgrade the data.

appscale/tools/appscale_tools.py
```diff
@@ -2,22 +2,25 @@
...
 import uuid
-
+import yaml
 
 # AppScale-specific imports
 from agents.factory import InfrastructureAgentFactory
 from appcontroller_client import AppControllerClient
 from appengine_helper import AppEngineHelper
 from appscale_logger import AppScaleLogger
+import appscale
```
cdonati: Is this import used anywhere?

appscale/tools/appscale_tools.py
```diff
@@ -62,11 +66,39 @@ class AppScaleTools(object):
   # A regular expression that matches files compressed in the zip format.
   ZIP_REGEX = re.compile(r'.zip\Z')
 
+
   # A str that contains all of the authorizations that an AppScale cloud
   # administrator should be granted.
   ADMIN_CAPABILITIES = "upload_app"
 
 
+  # AppScale repository location on local filesystem.
```
cdonati: Perhaps "on an AppScale image" rather than "local filesystem".

appscale/tools/appscale_tools.py
```diff
@@ -62,11 +66,39 @@ class AppScaleTools(object):
...
+  APPSCALE_REPO = "~/appscale"
+
+
+  # Command to run the Bootstrap.
+  RUN_BOOTSTRAP_COMMAND = "bash bootstrap_force_upgrade.sh --local-ip"
+
+
+  # String of commands to cd into the AppScale repository and run the Bootstrap.
+  BOOTSTRAP_COMMAND = "cd " + APPSCALE_REPO + ";" + RUN_BOOTSTRAP_COMMAND
```
cdonati: When it's something this specific, I don't think it makes sense for it to be defined as a constant. You can just construct it as needed below.

appscale/tools/appscale_tools.py
```diff
@@ -62,11 +66,39 @@ class AppScaleTools(object):
...
+
+
+  # Command to run the upgrade script from /appscale/scripts directory.
+  UPGRADE_SCRIPT = "python " + APPSCALE_REPO + "/scripts/upgrade.py"
+
+
+  # Git command to get the latest tags for a repository.
+  GIT_COMMAND = "git ls-remote --tags https://github.com/AppScale/appscale.git"
```
cdonati: I would give this a more descriptive name so that it's more obvious what it is doing when used below. Perhaps `GIT_LIST_TAGS`.

appscale/tools/appscale_tools.py
```diff
@@ -62,11 +66,39 @@ class AppScaleTools(object):
...
+
+
+  # Location of the upgrade status file on the remote machine.
+  UPGRADE_STATUS_FILE_LOC = '/var/log/appscale/upgrade-status-'
+
+  # .JSON file extention
+  JSON_FILE_EXTENTION = ".json"
```
cdonati: This doesn't make sense to me to have as a constant. If you anticipate needing to change the extension for the status file in the future, you could define it here, but you'd probably want to call it something more descriptive.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
       shutil.rmtree(file_location)
 
     return (login_host, http_port)
+
+  @classmethod
+  def upgrade(cls, options):
+    """ Upgrades the deployment to the latest AppScale version as well as
+    updates the data.
```
cdonati: Same comment as far as not always updating data.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+    zk_ips = ips_layout_yaml['zookeeper']
+    db_ips = ips_layout_yaml['database']
+    master_ip = ips_layout_yaml['master']
+    upgrade_version_available = cls.get_upgrade_version_available(master_ip, options.keyname)
+
+    if APPSCALE_VERSION == upgrade_version_available:
+      AppScaleLogger.log("AppScale is already at its latest code version, "
+        "so skipping code pull and build.")
+      AppScaleLogger.log("Running upgrade script to check if any other upgrade is needed.")
+      cls.shut_down_appscale_if_running(options)
```
cdonati: What's the use case for running the upgrade script if the user is already using the latest version?

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+      cls.run_upgrade_script(options, upgrade_version_available, master_ip, zk_ips, db_ips)
+      return
+
+    cls.shut_down_appscale_if_running(options)
+    cls.upgrade_appscale_code(options)
+    cls.run_upgrade_script(options, upgrade_version_available, master_ip, zk_ips, db_ips)
```
cdonati: We may want to give a final success message to assure the user that all has gone well and they can re-start AppScale. Also, we should tell the user to upgrade their tools.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+    try:
+      RemoteHelper.ssh(master_ip, options.keyname, upgrade_script_command,
+        options.verbose)
+      upgrade_status_file = cls.UPGRADE_STATUS_FILE_LOC + timestamp + cls.JSON_FILE_EXTENTION
+      command = 'cat' + " " + upgrade_status_file
+      ssh_file = RemoteHelper.get_command_output_from_remote(master_ip, command, options.keyname)
+      upgrade_status = ssh_file.stdout.read()
+
+      json_status = json.loads(upgrade_status)
+      if not json_status.keys():
```
cdonati: We should probably have an explicit message for this in the status file rather than assume it is the case if there are no keys.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+        AppScaleLogger.log("AppScale is already at its latest version.")
+
+      for key in json_status.keys():
+        for second_level_key in json_status[key]:
+          if second_level_key == 'Completion Status' and \
```
cdonati: Typically, JSON keys do not have spaces in them. Also, you don't need to loop through a dict if you just need one key.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+            json_status[key][second_level_key] == 'Success':
+            AppScaleLogger.success("{} process was successfully executed.".format(key))
+          elif not json_status[key][second_level_key] == 'Success':
```
cdonati: I don't think this is doing what you intend it to. This is checking the content of all keys that are _not_ 'Completion Status'.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+            AppScaleLogger.warn("Error encountered during upgrade process -> {} : {}".
+              format(second_level_key, json_status[key][second_level_key]))
+            AppScaleLogger.warn("For more information refer to " + upgrade_status_file
+              + " by logging into your head node.")
+
+    except ShellException:
+      AppScaleLogger.warn("Error executing upgrade script.")
+
+  @classmethod
+  def get_ip_str_for_command(cls, ip_list):
```
cdonati: I don't think you need this function. You should be able to achieve the same effect with `json.dumps(ip_list)`.

appscale/tools/appscale_tools.py
```diff
@@ -654,3 +686,145 @@ def upload_app(cls, options):
...
+
+    response = raw_input("Are you sure you want to proceed? (Y/N) ")
+    if response.lower() not in ['y', 'yes']:
+      raise AppScaleException("Cancelled AppScale upgrade")
+    else:
+      for ip in options.unique_ips:
+        try:
+          command = cls.BOOTSTRAP_COMMAND + " " + ip
+          RemoteHelper.ssh(ip, options.keyname, command, options.verbose)
+          AppScaleLogger.success("Successfully upgraded AppScale to its latest "
```
cdonati: The user may take this to mean that the full upgrade process is finished.

appscale/tools/parse_args.py
```diff
@@ -361,6 +361,13 @@ def add_allowed_flags(self, function):
         help="the name of the AppController property to set")
       self.parser.add_argument('--property_value',
         help="the value of the AppController property to set")
+    elif function == "appscale-upgrade":
+      self.parser.add_argument('--keyname', '-k', default=self.DEFAULT_KEYNAME,
+        help="the keypair name to use")
+      self.parser.add_argument('--ips_layout',
+        help="a YAML file dictating the placement strategy")
+      self.parser.add_argument('--unique_ips',
```
cdonati: Does this command need a list of unique IP addresses if it already has the full ips_layout?

appscale/tools/remote_helper.py
```diff
@@ -1042,3 +1042,16 @@ def collect_appcontroller_crashlog(cls, host, keyname, is_verbose):
       message = "AppController at {0} crashed for reasons unknown.".format(host)
 
     return message
+
+  @classmethod
+  def get_command_output_from_remote(cls, host, command, keyname, user='root', shell=False):
```
cdonati: Unless I'm missing something, you should be able to just use the `ssh` function in this class.

appscale/tools/appscale_tools.py
tmarballi: The problem with using json.dumps(list) or converting the list to string using any existing utilities is that it leaves spaces between list items. And because I am sending the list as an argument to the python script, it interprets the spaces as args too. Hence this function which builds the list without spaces.

appscale/tools/remote_helper.py
tmarballi: I had a problem using the existing ssh method while trying to call the command git ls remote to fetch all the tags, so I created this new method.

appscale/tools/appscale_tools.py
tmarballi: In the case where the user has manually updated the AppScale code for 3.0 (without using the upgrade tool command) but the data has not been upgraded.

appscale/tools/parse_args.py
tmarballi: For running the bootstrap on all machines, I needed the unique ips and there was an already existing method here to parse that from ips layout yaml. On the appscale-tools, I would have to parse it from a string, so I thought it would be better to pass it here.

bin/appscale
```diff
@@ -178,7 +178,10 @@ elif command in ["--version", "-v"]:
   sys.exit(0)
 elif command == "upgrade":
   try:
-    appscale.upgrade()
+    if len(sys.argv) == 3 and (sys.argv[2] == "--stash"):
+      appscale.upgrade(True)
```
menivaitsi: Change this to: 
`appscale.upgrade(stash=True)`

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/1583/

PR #508, opened 2016-05-27
Don't try to get the locale when generating a crashlog

cdonati:
The function can be problematic on some systems, and we don't
want to get in the way of generating a crashlog.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1518/

PR #509, opened 2016-06-02
Don't ask removal confirmation when test is true

cdonati:
This also checks the value of "test" when deploying an application.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1544/

PR #516, opened 2016-07-16
Recover AppController crashlog if init fails

cdonati:
This is intended to tell the user to run `appscale upgrade` if their data layout is too old.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1687/

PR #519, opened 2016-07-22
Use older version of wstools

cdonati:
This is a dependency for SOAPpy.
0.4.4 does not build with setuptools.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/1715/

PR #528, opened 2016-08-04
Specify clean, down, destroy commands a bit better.

obino:
Proposal to have clean, down, and destroy have slightly different meaning when running on a public cloud. Clean, would destroy instances and volumes, destroy would terminate the instances (but won't touch volumes), and down will simply stop appscale without terminating instances.


obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1963/

appscale/tools/appscale.py
```diff
@@ -740,11 +716,15 @@ def relocate(self, appid, http_port, https_port):
...
 
-  def destroy(self):
-    """ 'destroy' provides a nicer experience for users than the
+  def down(self, clean=False, terminate=False):
+    """ 'down' provides a nicer experience for users than the
     appscale-terminate-instances command, by using the configuration options
     present in the AppScalefile found in the current working directory.
 
+    Args:
+      terminate: A boolean to indicate if instances needs to be terminated
```
cdonati: Also document the `clean` argument.

appscale/tools/appscale.py
```diff
@@ -755,6 +735,19 @@ def destroy(self):
...
+      command.append("--verbose")
+    else:
+      is_verbose = False
+
+    if 'keyname' in contents_as_yaml:
+      keyname = contents_as_yaml['keyname']
+      command.append("--keyname")
+      command.append(contents_as_yaml['keyname'])
+    else:
+      keyname = 'appscale'
```
cdonati: Is this behavior done elsewhere? It would make more sense to me to raise an exception, but I might be missing something.

appscale/tools/appscale.py
```diff
@@ -764,12 +757,19 @@ def destroy(self):
     if "EC2_URL" in contents_as_yaml:
       os.environ["EC2_URL"] = contents_as_yaml["EC2_URL"]
 
-    if 'keyname' in contents_as_yaml:
-      command.append("--keyname")
-      command.append(contents_as_yaml['keyname'])
+    if clean:
+      LocalState.ensure_user_wants_to_clean()
```
cdonati: This should probably be wrapped with a check for `test` and maybe `force`.

appscale/tools/appscale.py
```diff
@@ -764,12 +757,19 @@ def destroy(self):
...
+      all_ips = LocalState.get_all_public_ips(keyname)
+      for ip in all_ips:
+        RemoteHelper.ssh(ip, keyname, self.TERMINATE, is_verbose)
+      AppScaleLogger.success("Successfully cleaned your AppScale deployment.")
 
-    if 'verbose' in contents_as_yaml and contents_as_yaml['verbose'] == True:
-      command.append("--verbose")
+    if terminate:
+      infrastructure = LocalState.get_infrastructure(keyname)
```
cdonati: Did you check to see what this returns if infrastructure is not defined?

appscale/tools/appscale.py
```diff
@@ -764,12 +757,19 @@ def destroy(self):
...
+      if infrastructure != "xen" and not LocalState.are_disks_used(
```
cdonati: Shouldn't the user still be able to terminate the instances if they have disks attached?

appscale/tools/appscale.py
```diff
@@ -764,12 +757,19 @@ def destroy(self):
...
+        keyname) and not 'test' in contents_as_yaml:
```
cdonati: `'test' not in contents_as_yaml` is more Pythonic.

appscale/tools/appscale_tools.py
```diff
@@ -604,28 +604,32 @@ def terminate_instances(cls, options):
       AppScaleException: If AppScale is not running, and thus can't be
       terminated.
     """
-    if not os.path.exists(LocalState.get_secret_key_location(options.keyname)):
-      raise AppScaleException("AppScale is not running with the keyname {0}".
+    try:
+      infrastructure = LocalState.get_infrastructure(options.keyname)
```
cdonati: Same comment here concerning if infrastructure is not defined.

appscale/tools/appscale_tools.py
```diff
@@ -604,28 +604,32 @@ def terminate_instances(cls, options):
...
-    infrastructure = LocalState.get_infrastructure(options.keyname)
-
-    # If the user is on a cloud deployment, and not backing their data to
-    # persistent disks, warn them before shutting down AppScale.
-    # Also, if we're in developer mode, skip the warning.
-    if infrastructure != "xen" and not LocalState.are_disks_used(
-      options.keyname) and not options.test:
-      LocalState.ensure_user_wants_to_terminate()
+    if infrastructure == "xen" or not options.terminate:
+      # We are in cluster mode: let's chek AppScale is running.
```
cdonati: chek -> check if

appscale/tools/appscale_tools.py
```diff
@@ -604,28 +604,32 @@ def terminate_instances(cls, options):
...
 
-    if infrastructure in InfrastructureAgentFactory.VALID_AGENTS:
-      RemoteHelper.terminate_cloud_infrastructure(options.keyname,
-        options.verbose)
-    else:
+    # Stop gracefully the AppScale deployment.
+    try:
       RemoteHelper.terminate_virtualized_cluster(options.keyname,
         options.verbose)
+    except:
```
cdonati: Please determine what exceptions can be thrown here. I don't think it makes sense to give a success message if the tools fail to stop services.

appscale/tools/appscale_tools.py
```diff
@@ -604,28 +604,32 @@ def terminate_instances(cls, options):
...
+      # Don't fail if we cannot find the configuration.
+      pass
 
-    LocalState.cleanup_appscale_files(options.keyname)
-    AppScaleLogger.success("Successfully shut down your AppScale deployment.")
```
cdonati: What's the output on a cluster? Does the user get any feedback on whether their `down` completed successfully?

appscale/tools/appscale_tools.py
```diff
@@ -604,28 +604,32 @@ def terminate_instances(cls, options):
...
+    # And if we are on a cloud infrasrtucture, terminate instances if
```
cdonati: infrastructure

appscale/tools/local_state.py
```diff
@@ -93,9 +92,24 @@ def ensure_appscale_isnt_running(cls, keyname, force):
...
+        login_host = cls.get_login_host(keyname)
+        secret_key = cls.get_secret_key(keyname)
+      except (IOError, AppScaleException, BadConfigurationException):
+        # If we don't have the locations files, we are not running.
+        return
+
+      acc = AppControllerClient(login_host, secret_key)
+      try:
+        acc.get_status()
+      except Exception:
```
cdonati: It would be nice if this were more specific.

appscale/tools/local_state.py
```diff
@@ -705,16 +719,17 @@ def get_oauth2_storage_location(cls, keyname):
...
       keyname: The SSH keypair name that uniquely identifies this AppScale
         deployment.
     """
-    files_to_remove = [LocalState.get_locations_yaml_location(keyname),
-      LocalState.get_locations_json_location(keyname),
-      LocalState.get_secret_key_location(keyname)]
+    files_to_remove = [LocalState.get_secret_key_location(keyname)]
+    if remove_location:
+      files_to_remove += [LocalState.get_locations_yaml_location(keyname),
+      LocalState.get_locations_json_location(keyname)]
```
cdonati: Use vertical alignment or add a line break after the opening bracket.

appscale/tools/local_state.py
```diff
@@ -965,15 +980,15 @@ def generate_crash_log(cls, exception, stacktrace):
...
+  def ensure_user_wants_to_clean(cls):
+    """ Asks the user for confirmation before we clean their AppScale
     deployment.
 
     Raises:
       AppScaleException: If the user does not want to terminate their
         AppScale deployment.
     """
-    cls.confirm_or_abort("Terminating AppScale will delete all stored data.")
+    cls.confirm_or_abort("Cleaning or terminating AppScale will delete all stored data.")
```
cdonati: I think we lose some specificity here. If the user is just doing a `clean`, it might be alarming to see a message that mentions terminate.

appscale/tools/remote_helper.py
```diff
@@ -122,7 +121,24 @@ def start_head_node(cls, options, my_id, node_layout):
...
     if options.infrastructure:
-      instance_id, public_ip, private_ip = cls.spawn_node_in_cloud(options)
+      agent = InfrastructureAgentFactory.create_agent(options.infrastructure)
+
+      params = agent.get_params_from_args(options)
+      public_ips, private_ips, instance_ids = agent.describe_instances(params)
+      try:
+        login_ip = LocalState.get_login_host(options.keyname)
+      except (IOError, BadConfigurationException):
+        # We don't have the location file, we need to start instances.
```
cdonati: What do we do with running instances in this case?

appscale/tools/remote_helper.py
```diff
@@ -122,7 +121,24 @@ def start_head_node(cls, options, my_id, node_layout):
...
+        login_ip = None
+
+      if login_ip in public_ips:
```
cdonati: I would be a little more defensive about this rather than relying on `describe_instances` to never return `None` in the list of `public_ips`.

appscale/tools/remote_helper.py
```diff
@@ -122,7 +121,24 @@ def start_head_node(cls, options, my_id, node_layout):
...
+        index = public_ips.index(login_ip)
+        public_ip = public_ips[index]
```
cdonati: This declaration is never used.

appscale/tools/remote_helper.py
```diff
@@ -825,7 +841,7 @@ def terminate_cloud_infrastructure(cls, keyname, is_verbose):
       is_verbose: A bool that indicates if we should print the commands executed
         to stdout.
     """
-    AppScaleLogger.log("About to terminate instances spawned with keyname {0}"
+    AppScaleLogger.log("Will terminate deployment and instances spawned with keyname {0}"
```
cdonati: Since it seems like the reason for this message is to give the user an opportunity to cancel, we may want to mention that they can Ctrl-C.

appscale/tools/remote_helper.py
```diff
@@ -889,12 +907,18 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
...
-    AppScaleLogger.log("Terminating instances in a virtualized cluster with " +
-      "keyname {0}".format(keyname))
+    AppScaleLogger.log("Terminating appscale deployment with keyname" +
+      " keyname {0}".format(keyname))
     time.sleep(2)
 
     shadow_host = LocalState.get_host_with_role(keyname, 'shadow')
-    acc = AppControllerClient(shadow_host, LocalState.get_secret_key(keyname))
+    try:
+      acc = AppControllerClient(shadow_host, LocalState.get_secret_key(keyname))
```
cdonati: I would put the try/except around the LocalState call so that it's clear which call is throwing the exception.

appscale/tools/remote_helper.py
```diff
@@ -889,12 +907,18 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
...
+    except IOError:
+      # We couldn't find the secret key: AppScale is most likely not
+      # running.
+      AppScaleLogger.log("Couldn't find AppScale secret key.")
```
cdonati: Should this raise an exception so that the tools return a non-zero exit code?

appscale/tools/appscale.py
```diff
@@ -764,12 +757,19 @@ def destroy(self):
...
       os.environ["EC2_URL"] = contents_as_yaml["EC2_URL"]
 
-    if 'keyname' in contents_as_yaml:
-      command.append("--keyname")
-      command.append(contents_as_yaml['keyname'])
+    if clean:
+      LocalState.ensure_user_wants_to_clean()
+      all_ips = LocalState.get_all_public_ips(keyname)
+      for ip in all_ips:
+        RemoteHelper.ssh(ip, keyname, self.TERMINATE, is_verbose)
```
cdonati: This kills the AppController, and `terminate_virtualized_cluster` later tries to talk to the AC, which results in an unnecessary warning log.

appscale/tools/appscale.py
obino: Yes I have seen this behavior before. I assume is part of the cluster mode code path.

appscale/tools/appscale.py
obino: the tools will define it no matter what (ie locations.yaml has an infrasrtucture field). If the file doesn't exists we'll get an exception.

appscale/tools/appscale.py
obino: Yes, the check is to ensure that if we terminate an ephemeral instance (terminate == true), the user is aware that data can be lost if no volume has been used.

appscale/tools/appscale_tools.py
obino: The output is now part of the command in appscale.py.  I moved it there since we have also similar output for clean.

appscale/tools/local_state.py
obino: Would you prefer to have 2 functions then with 2 different messages? Or should we phrase is differently, say The following operation will delete or may delete ...?

appscale/tools/remote_helper.py
obino: If we don't have the location file the instances are effectively lost. We could in theory find them running under the proper keyname,  but we wouldn't know the role of each one of them.

appscale/tools/appscale.py
obino: Good point. What would you suggest to do here? the terminate.rb clean is needed to clean the disk. Should I check if clean is set in terminate_virtualized_instances and avoid the call there?

obino:
ocd run https://ocd.appscale.com:8080/job/Daily%20Build/1988/

obino:
OCD run https://ocd.appscale.com:8080/job/Daily%20Build/1994/

menivaitsi:
Refer to https://github.com/AppScale/appscale-tools/pull/559

PR #534, opened 2016-09-12
Catch struct.error when importing AzureAgent

cdonati:
Currently, msrestazure 0.4.2 is not working properly on Precise.
```
>>> import msrestazure.azure_active_directory
struct.error: unpack requires a string argument of length 4
```


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2018/

PR #536, opened 2016-09-23
Require a specific version of adal

cdonati:
0.4.1 does not install correctly without jwt.


cdonati:
https://ocd.appscale.com:8080/job/Tools-test/988/

PR #546, opened 2016-11-05
Remove version check from setup

cdonati:
Existing packages in the appscale namespace may prevent importing from uninstalled tools package. Users will still receive a warning at runtime if they have a Python version older than 2.6.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2203/

PR #548, opened 2016-11-10
Replace scripts with entry points

cdonati:
This avoids installing the tools package as an egg. Egg packages have trouble co-existing with non-egg packages that are in the same namespace.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2234/

menivaitsi:
Can we do an upgrade demo from current master on this one?

PR #557, opened 2016-12-03
Set root SSH key for GCE instances via CommonInstaceMetadata

menivaitsi:


menivaitsi:
See https://github.com/AppScale/appscale-tools/pull/558 instead

PR #560, opened 2016-12-08
terminate cloud instances when tools cannot ssh into private ips

whoarethebritons:


whoarethebritons:
no longer a relevant fix

PR #561, opened 2016-12-08
Use yaml.safe_load to get rid of unicode string issues

whoarethebritons:


whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2371/

PR #564, opened 2016-12-21
Azure scale sets

menivaitsi:
From Tanvi's branch: https://github.com/AppScale/appscale-tools/pull/562
Build: https://ocd.appscale.com:8080/job/Daily%20Build/2414/


appscale/tools/remote_helper.py
```diff
@@ -769,16 +769,13 @@ def wait_for_machines_to_finish_loading(cls, host, keyname):
         deployment.
       keyname: The name of the SSH keypair used for this AppScale deployment.
     """
-    acc = AppControllerClient(host, LocalState.get_secret_key(keyname))
-    all_ips = acc.get_all_public_ips()
+    while True:
+      acc = AppControllerClient(host, LocalState.get_secret_key(keyname))
```
cdonati: This can be created before the while loop.

appscale/tools/appscale_tools.py
```diff
@@ -505,14 +505,26 @@ def run_instances(cls, options):
       AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
                          "for use. This can take few minutes.")
 
+      head_node.public_ip = public_ips[0]
```
cdonati: Is this change needed because only the first node gets a public IP address in Azure? Since the `start_all_nodes` function does not imply this, I think an explanatory comment would be helpful here.

appscale/tools/appscale_tools.py
```diff
@@ -505,14 +505,26 @@ def run_instances(cls, options):
...
+      # and another for scale set(s).
+      node_layout.nodes.remove(head_node)
+
       # Set newly obtained node layout info for this deployment.
-      for i, _ in enumerate(instance_ids):
-        node_layout.nodes[i].public_ip = public_ips[i]
-        node_layout.nodes[i].private_ip = private_ips[i]
-        node_layout.nodes[i].instance_id = instance_ids[i]
+      for i, _ in enumerate(instance_ids[1:]):
+        if "shadow" in node_layout.nodes[i].roles:
```
cdonati: What's the reason for skipping this?

appscale/tools/appscale_tools.py
```diff
@@ -505,14 +505,26 @@ def run_instances(cls, options):
...
+          continue
+        node_layout.nodes[i].public_ip = public_ips[i+1]
+        node_layout.nodes[i].private_ip = private_ips[i+1]
+        node_layout.nodes[i].instance_id = instance_ids[i+1]
```
cdonati: It seems to me that this could be simplified a bit if you loop on nodes rather than instance_ids. For example, instead of removing the head node and then re-adding it later, you could do
```
for index, node in enumerate(node_layout.nodes):
  node.public_ip = public_ips[index]
  node.private_ip = private_ips[index]
  node.instance_id = instance_ids[index]
```

appscale/tools/agents/azure_agent.py
```diff
@@ -226,7 +244,9 @@ def describe_instances(self, parameters, pending=False):
     credentials = self.open_connection(parameters)
     subscription_id = parameters[self.PARAM_SUBSCRIBER_ID]
     resource_group = parameters[self.PARAM_RESOURCE_GROUP]
-    network_client = NetworkManagementClient(credentials, subscription_id)
+
+    network_client = NetworkManagementClient(credentials, subscription_id,
+                                             api_version='2016-09-01')
```
cdonati: I would recommend creating a constant for this.

appscale/tools/agents/azure_agent.py
```diff
@@ -351,9 +376,142 @@ def create_virtual_machine(self, credentials, network_client, network_id,
...
+    # Create a scale set using the count of VMs provided.
+    else:
+      scale_set_name = random_resource_name + "-scaleset-{}vms".format(count)
+      AppScaleLogger.verbose('Creating a Scale Set {0} with {1} VMs'.
+                             format(scale_set_name, count), verbose)
+      self.create_scale_set(count, parameters, random_resource_name,
+                            scale_set_name, subnet)
+
+  def create_scale_set(self, count, parameters, resource_name,
+                       scale_set_name, subnet,):
```
cdonati: trailing comma

appscale/tools/appscale_tools.py
menivaitsi: At this time we have no good way of knowing the position of the head node in node_layout.nodes.
Since the Azure agent only returns a public IP for the head node, we are updating that first with a different accessor (head_node) and then we update the rest of the list with just private IPs.

The real solution would be to modify the NodeLayout parser to deterministically put the head node(s)/loadbalancer(s) first in the list. That was too much work at the time.

appscale/tools/appscale_tools.py
menivaitsi: See comment above.

appscale/tools/appscale_tools.py
menivaitsi: We need to skip the head node here. Is there a way to enumerate part of the list and not the entire one?

appscale/tools/appscale_tools.py
cdonati: Sorry, I still don't understand. If the head node is the shadow node, then it has already been removed from `node_layout.nodes`, so this would never get hit. If there are multiple shadow nodes, then wouldn't this prevent other nodes with that role from being assigned an IP address?

appscale/tools/appscale_tools.py
cdonati: Since the head node is already being popped and re-inserted from the list, why not just do it here to ensure it's first?
```
node_layout.nodes.remove(head_node)
node_layout.nodes.insert(0, head_node)
[iterate over entire node list]
```

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/2491/

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2497/

Issue #565, opened 2017-01-03
Error in running appscale.bash


khaledkbadr:
When I run `bash appscale-tools/debian/appscale_build.sh`, It seems that there's requests version conflict in the script
```
Installed /usr/local/lib/python2.7/dist-packages/appscale_tools-3.1.0-py2.7.egg
Processing dependencies for appscale-tools==3.1.0
Searching for requests>=2.7.0
Reading https://pypi.python.org/simple/requests/
Downloading https://pypi.python.org/packages/5b/0b/34be574b1ec997247796e5d516f3a6b6509c4e064f2885a96ed885ce7579/requests-2.12.4.tar.gz#md5=acdb48888a9d3c7309da12fc7f83fedb
Best match: requests 2.12.4
Processing requests-2.12.4.tar.gz
Writing /tmp/easy_install-SQy_XG/requests-2.12.4/setup.cfg
Running requests-2.12.4/setup.py -q bdist_egg --dist-dir /tmp/easy_install-SQy_XG/requests-2.12.4/egg-dist-tmp-JaMIKU
warning: no files found matching 'test_requests.py'
creating /usr/local/lib/python2.7/dist-packages/requests-2.12.4-py2.7.egg
Extracting requests-2.12.4-py2.7.egg to /usr/local/lib/python2.7/dist-packages
Adding requests 2.12.4 to easy-install.pth file

Installed /usr/local/lib/python2.7/dist-packages/requests-2.12.4-py2.7.egg
error: requests 2.12.4 is installed but requests!=2.12.*,>=2.0.0 is required by set(['adal'])
Unable to complete AppScale tools installation.

```

PR #566, opened 2017-01-03
Fix requests version

khaledkbadr:
#565


cdonati:
Hi @khaledkbadr. Thank you for the pull request. Unfortunately, it appears that this syntax is not supported by versions of pip that are older than 6. I want to check if there's an alternative solution that we can implement that will maintain support for older pip installations.

It doesn't look like requests is imported directly by appscale-tools, so we may get away with removing it from the setup.py file altogether. Would you mind trying that to see if it fixes your issue?

khaledkbadr:
I did fix my issue by cloning the repo and running the bash script directly.
That error happened when following installing from source in [try-appscale](https://www.appscale.com/try-appscale/)

But when I just cloned the repos and installed them manually it works smoothly.

cdonati:
Okay, the bootstrap script uses the `3.1.0` tag from the git repository. I believe the issue there was that `adal` was not locked to 4.0.0 for `appscale-tools 3.1.0`. Since the issue has been fixed in the latest master, you can use `bash ./bootstrap.sh --tag dev`.

We plan to have a new release within the next month, so the vanilla bootstrap should be fixed at that point.

PR #568, opened 2017-01-11
Init templates

menivaitsi:


appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,160 @@
...
+#   - [vm3_private_IP | node-3]
+#  database:   # Cassandra needs a quorum to answer queries (hence 3)
+#   - [vm4_private_IP | node-4]
+#   - [vm5_private_IP | node-5]
+#   - [vm6_private_IP | node-6]
+#  zookeeper:  # Zookeeper needs a majority of peers online (hence 3)
+#   - [vm7_private_IP | node-7]
+#   - [vm8_private_IP | node-8]
+#   - [vm9_private_IP | node-9]
+#login : <vm1_public_IP>
```
cdonati: Missing a space. Also, should we mention that DNS can also be used?

appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,160 @@
...
+# Use the IPs of existing machines,
+# or node-X notation if you want AppScale to start new machines on the
+# Infrastructure of choice.
+# The following layout provides Fault Tolerance and High Availability:
+# ips_layout:
+#  master : [vm1_private_IP | node-1]
+#  appengine:
+#   - [vm2_private_IP | node-2]
+#   - [vm3_private_IP | node-3]
+#  database:   # Cassandra needs a quorum to answer queries (hence 3)
```
cdonati: The 'hence 3' is confusing here since `replication` is set to 1 by default.

menivaitsi: If we have replication set to 3 by default, what would be the behavior when starting AppScale with 1 database instead of 3?

Would that make more sense since we propose 3 DBs at all times?

appscale/tools/appscale.py
```diff
@@ -85,7 +92,7 @@ class AppScale():
   get <regex>                       Gets all AppController properties matching
                                     the provided regex: for developers only.
   help                              Displays this message.
-  init <cloud|cluster>              Writes a new configuration file for
+  init [cloud|cluster]              Writes a new configuration file for
```
obino: Do we need the cloud|cluster at all? WHat about we remove it from the description, but we still tolerate for backward compatibility?

menivaitsi: This branch does not remove any of the previous templates or functionality. 
I believe it should be a separate pull after 3.2.

appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,162 @@
...
+   - [vm3_private_IP | node-3]
+  database:   # Cassandra needs a quorum to answer queries (hence 3)
+   - [vm4_private_IP | node-4]
+   - [vm5_private_IP | node-5]
+   - [vm6_private_IP | node-6]
+  zookeeper:  # Zookeeper needs a majority of peers online (hence 3)
+   - [vm7_private_IP | node-7]
+   - [vm8_private_IP | node-8]
+   - [vm9_private_IP | node-9]
+login : [vm1_public_IP | Domain Name]
```
obino: I think we should have some comment for login. Something like: Login is used to instruct AppScale which one is the visible public IP used by clients to connect. It could be the public IP of the load-balancer instance, or it could be the list of IPs used for exernal load-balancers.

appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,162 @@
...
+#################################################
+########## Public Cloud configuration ###########
+#################################################
+
+############ Amazon Web Services EC2 ############
+# To deploy on AWS EC2, uncomment/fill out the following parameters:
+# infrastructure : 'ec2'
+
+# AppScale-ready AWS AMI.
+# machine : 'ami-XXXXXX’
```
obino: Usually I see the comment having a space after the #, but the command not: it simplifies the activation of the command (a single chracter to cancel :D).

appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,162 @@
...
+
+# The availability zone that should be used for all virtual machines in your
+# AppScale deployment. If using persistent disks in your AppScale deployment,
+# they should be in the same zone that you specify here.
+# zone : ‘West-US’
+
+# The unique resource group that Azure resources will be created under for this
+# deployment.
+# By default, the group is auto-generated.
+# group : 'appscalegroup'
```
obino: Group is common to all the clouds, so I propose to have it only once in the below optional paramenters. It's rarely used anyway.

appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,162 @@
...
+# azure_tenant_id : ‘XXXX’
+# azure_app_secret_key : ‘YYYY’
+
+# Instance type of your choice. Default is Standard_A3.
+# instance_type : ‘Standard_A3’
+
+# The availability zone that should be used for all virtual machines in your
+# AppScale deployment. If using persistent disks in your AppScale deployment,
+# they should be in the same zone that you specify here.
+# zone : ‘West-US’
```
tmarballi: I think the zone for Azure is without the dash i.e. West US. I have not tested with the dash to be sure it does not work but the available zones mentioned in their SDK documentation is without dashes.

appscale/tools/templates/AppScalefile
tmarballi: The resource group parameter to use for Azure is named as 'azure_resource_group'. This 'group' is the common security group name used for the rest of the cloud platforms.

appscale/tools/templates/AppScalefile
```diff
@@ -0,0 +1,162 @@
...
+
+# The unique resource group that Azure resources will be created under for this
+# deployment.
+# By default, the group is auto-generated.
+# group : 'appscalegroup'
+
```
tmarballi: The 'azure_storage_account' parameter is missing which is also an optional field unless provided in the AppScalefile, then we use that one which is mostly the case.

menivaitsi: Thanks! These changes were already underway but you're faster :)

appscale/tools/templates/AppScalefile
menivaitsi: I believe YAML parsers remove leading whitespaces and this should not be a problem.
It's a lot of info and gets clumped together without spaces.

PR #569, opened 2017-01-18
Load balancer distinction

whoarethebritons:
requires #564

### DIFF
--- a/appscale/tools/remote_helper.py
+++ b/appscale/tools/remote_helper.py
@@ -96,7 +96,7 @@ class RemoteHelper(object):
 
 
   @classmethod
-  def start_all_nodes(cls, options, count):
+  def start_all_nodes(cls, options, node_layout):
     """ Starts all nodes in the designated public cloud.
 
     Args:
@@ -108,10 +108,62 @@ class RemoteHelper(object):
       The public IPs and instance IDs (dummy values in non-cloud deployments)
       corresponding to the nodes that were started.
     """
-    instance_ids, public_ips, private_ips = cls.spawn_nodes_in_cloud(
-      options, count=count)
+    agent = InfrastructureAgentFactory.create_agent(options.infrastructure)
+    params = agent.get_params_from_args(options)
 
-    return instance_ids, public_ips, private_ips
+    # If we have running instances under the current keyname, we try to
+    # re-attach to them. If we have issue finding the locations file or the
+    # IP of the head node, we throw an exception.
+    login_ip = None
+    public_ips, private_ips, instance_ids = agent.describe_instances(params)
+    if public_ips:
+      try:
+        login_ip = LocalState.get_login_host(options.keyname)
+      except (IOError, BadConfigurationException):
+        raise AppScaleException(
+          "Couldn't get login ip for running deployment with keyname"
+          " {}.".format(options.keyname))
+      if login_ip not in public_ips:
+        raise AppScaleException(
+          "Couldn't recognize running instances for deployment with"
+          " keyname {}.".format(options.keyname))
+
+    if login_ip in public_ips:
+      AppScaleLogger.log("Reusing already running instances.")
+      # Set newly obtained node layout info for this deployment.
+      for i, _ in enumerate(instance_ids):
+        node_layout.nodes[i].public_ip = public_ips[i]
+        node_layout.nodes[i].private_ip = private_ips[i]
+        node_layout.nodes[i].instance_id = instance_ids[i]
+
+      return node_layout
+
+    agent.configure_instance_security(params)
+
+    instance_ids, public_ips, private_ips = \
+      cls.spawn_load_balancers_in_cloud(options, agent, params,
+                                        len(node_layout.get_nodes(
+                                            'load_balancer', True)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
+                       "for use. This can take few minutes.")
+
+    o_instance_ids, o_public_ips, o_private_ips = \
+      cls.spawn_other_nodes_in_cloud(options, agent, params,
+                                              len(node_layout.get_nodes(
+                                                'load_balancer', False)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
+                       "for use. This can take few minutes.")
+
+    instance_ids.extend(o_instance_ids)
+    public_ips.extend(o_public_ips)
+    private_ips.extend(o_private_ips)
+    # Set newly obtained node layout info for this deployment.
+    for i, _ in enumerate(instance_ids):
+      node_layout.nodes[i].public_ip = public_ips[i]
+      node_layout.nodes[i].private_ip = private_ips[i]
+      node_layout.nodes[i].instance_id = instance_ids[i]
+
+    return node_layout
 
   @classmethod
   def enable_root_ssh(cls, options, public_ip):
@@ -210,7 +262,7 @@ class RemoteHelper(object):
 
 
   @classmethod
-  def spawn_nodes_in_cloud(cls, options, count=1):
+  def spawn_load_balancers_in_cloud(cls, options, agent, params, count=1):
     """Starts a single virtual machine in a cloud infrastructure.
 
     This method also prepares the virual machine for use by the AppScale Tools.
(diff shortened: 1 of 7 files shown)


appscale/tools/remote_helper.py
```diff
@@ -259,6 +284,25 @@ def spawn_nodes_in_cloud(cls, options, count=1):
 
 
   @classmethod
+  def spawn_other_nodes_in_cloud(cls, options, agent, params, count=1):
+    """Starts a single virtual machine in a cloud infrastructure.
```
tmarballi: I think the description here should be that it starts 'count' number of virtual machines and not just a single virtual machine.

appscale/tools/remote_helper.py
```diff
@@ -259,6 +284,25 @@ def spawn_nodes_in_cloud(cls, options, count=1):
...
+
+    This method also prepares the virual machine for use by the AppScale Tools.
```
tmarballi: virual -> virtual and the same on top for the spawn_load_balancers_in_cloud method (Line 266)

appscale/tools/remote_helper.py
```diff
@@ -259,6 +284,25 @@ def spawn_nodes_in_cloud(cls, options, count=1):
...
+    Args:
+      options: A Namespace that specifies the cloud infrastructure to use, as
+        well as how to interact with that cloud.
+      count: A int, the number of instances to start.
+    Returns:
+      The instance ID, public IP address, and private IP address of the machine
+        that was started.
+    """
+    instance_ids, public_ips, private_ips = agent.run_instances(count=count,
+                                                    parameters=params,
```
tmarballi: Could you please align these parameters vertically below the first one.

appscale/tools/remote_helper.py
```diff
@@ -96,22 +96,72 @@ class RemoteHelper(object):
...
+                                            'load_balancer', True)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
+                       "for use. This can take few minutes.")
+
+    if len(node_layout.get_nodes('load_balancer', False)) > 0:
+      instance_ids, public_ips, private_ips = \
+        cls.spawn_other_nodes_in_cloud(options, agent, params,
+                                      len(node_layout.get_nodes(
+                                          'load_balancer', False)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
```
tmarballi: Maybe remove this log line or change the description as the same line is present above.

appscale/tools/remote_helper.py
```diff
@@ -96,22 +96,72 @@ class RemoteHelper(object):
...
+
+    instance_ids, public_ips, private_ips = \
+      cls.spawn_load_balancers_in_cloud(options, agent, params,
+                                        len(node_layout.get_nodes(
+                                            'load_balancer', True)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
+                       "for use. This can take few minutes.")
+
+    if len(node_layout.get_nodes('load_balancer', False)) > 0:
+      instance_ids, public_ips, private_ips = \
```
cdonati: I think you need to extend your list of existing instances rather than discarding them. Otherwise, your load balancers will not get assigned IP addresses.

appscale/tools/remote_helper.py
```diff
@@ -96,22 +96,72 @@ class RemoteHelper(object):
...
+          " {}.".format(options.keyname))
+      if login_ip not in public_ips:
+        raise AppScaleException(
+          "Couldn't recognize running instances for deployment with"
+          " keyname {}.".format(options.keyname))
+
+    if login_ip in public_ips:
+      AppScaleLogger.log("Reusing already running instances.")
+      # Set newly obtained node layout info for this deployment.
+      for i, _ in enumerate(instance_ids):
```
menivaitsi: How do we know in this loop if we're assigning the public IP/private IP/instance ID to the correct node?
What if describe_instances does not return the IPs in the same order each time?

Can we use the locations mapping to assign private IP (and corresponding public IP, instance ID) to the proper node in node_layout?

appscale/tools/remote_helper.py
```diff
@@ -96,22 +96,72 @@ class RemoteHelper(object):
...
+        node_layout.nodes[i].public_ip = public_ips[i]
+        node_layout.nodes[i].private_ip = private_ips[i]
+        node_layout.nodes[i].instance_id = instance_ids[i]
+
+      return node_layout
+
+    agent.configure_instance_security(params)
+
+    instance_ids, public_ips, private_ips = \
```
menivaitsi: Please rewrite this to something like:
```
     instance_ids, public_ips, private_ips = cls.spawn_load_balancers_in_cloud(

       options, agent, params,

       len(node_layout.get_nodes('load_balancer', True)))
```
to be more readable

appscale/tools/remote_helper.py
```diff
@@ -96,22 +96,72 @@ class RemoteHelper(object):
...
+      cls.spawn_load_balancers_in_cloud(options, agent, params,
+                                        len(node_layout.get_nodes(
+                                            'load_balancer', True)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
+                       "for use. This can take few minutes.")
+
+    if len(node_layout.get_nodes('load_balancer', False)) > 0:
+      instance_ids, public_ips, private_ips = \
+        cls.spawn_other_nodes_in_cloud(options, agent, params,
```
menivaitsi: Same here about alignment.

appscale/tools/remote_helper.py
```diff
@@ -96,22 +96,72 @@ class RemoteHelper(object):
...
+                                      len(node_layout.get_nodes(
+                                          'load_balancer', False)))
+    AppScaleLogger.log("\nPlease wait for AppScale to prepare your machines "
+                       "for use. This can take few minutes.")
+
+    # Set newly obtained node layout info for this deployment.
+    for i, _ in enumerate(instance_ids):
```
menivaitsi: Should there be two loops; one for loadbalancers and one for the rest of nodes?
Alternatively, you should append to the existing lists of IPs/IDs the second batch you generate.

appscale/tools/remote_helper.py
whoarethebritons: I had previously had that but it wasn't working on azure because it was adding duplicates so discarding them worked because the describe instances returned from the agent returned all of the instances.

appscale/tools/remote_helper.py
whoarethebritons: I had previously had that but it wasn't working on azure because it was adding duplicates so discarding them worked because the describe instances returned from the agent returned all of the instances.

menivaitsi: @whoarethebritons what is the behavior for the other agents?

We need a single approach for all of them at this level.

appscale/tools/remote_helper.py
whoarethebritons: I believe this loop was added by Graziano when he was having nodes reattach... however that is a good point I will look into it.

whoarethebritons: @menivaitsi wait will this matter? the locations.json gets rewritten and those nodes shouldn't have anything "appscale" running. right?

menivaitsi: @whoarethebritons the services might be down but the data is in certain machines only.

Imagine a scenario where the database gets assigned a different IP. Suddenly, there's no data and it's a fresh Cassandra node.

I suggest we put some effort here to handle this case and test it.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2534/

appscale/tools/remote_helper.py
whoarethebritons: @menivaitsi I want to talk more about this with the team because I think it might not be a very simple thing to do... we cannot match up nodes from the locations.json to the node layout because the node layout will not have any of the information that can be matched besides roles which may have changed since the last up. I agree that this is a case that needs to look into but I think it might be out of scope for this PR.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2572/

PR #570, opened 2017-01-22
Include extra dependencies in app tarball

cdonati:
This uses the go-app-builder binary to discover the application's dependencies. The user must have a suitable `GOROOT` (that includes `bin/go-app-builder`) and `GOPATH` defined as environment variables in order for it to work.

If there is an issue discovering the required dependencies, the user will be prompted if they want to continue.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2541/

PR #572, opened 2017-01-23
Prevent setuptools 34 from being installed

cdonati:


debian/appscale_build.sh
```diff
@@ -17,12 +17,14 @@ if [ -f ./debian/control.${DIST} ]; then
     fi
 fi
 
+# If setuptools 34 has already been installed, subsequent commands
```
obino: it seems like the comment is not finished.

cdonati:
https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/3460/

PR #574, opened 2017-01-26
Azure agent returns public IPs only for load balancers

tmarballi:
This PR has added an additional flag for agents to return public IPs only if needed for load balancer role. Azure agent is the only one currently to implement that functionality and rest of the agents will follow later.


tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/2589/

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/2624/

PR #577, opened 2017-01-30
Terminate runs on head node

whoarethebritons:
requires AppScale/appscale#2297

### DIFF
--- a/appscale/tools/appcontroller_client.py
+++ b/appscale/tools/appcontroller_client.py
@@ -228,8 +228,7 @@ class AppControllerClient():
       self.server.start_roles_on_nodes, roles_to_nodes, self.secret)
 
   def is_appscale_terminated(self):
-    """Dynamically adds the given machines to an AppScale deployment, with the
-    specified roles.
+    """Queries the AppController to see if the system has been terminated
 
     Returns:
       A boolean indicating whether appscale has finished running terminate
@@ -240,8 +239,7 @@ class AppControllerClient():
                                  self.server.is_appscale_terminated, self.secret)
 
   def run_terminate(self, clean):
-    """Dynamically adds the given machines to an AppScale deployment, with the
-    specified roles.
+    """Tells the AppController to terminate AppScale on the deployment
 
     Args:
       clean: A boolean indicating whether the clean parameter should be
@@ -251,16 +249,13 @@ class AppControllerClient():
     """
     return self.server.run_terminate(clean, self.secret)
 
-
   def receive_server_message(self):
-    """Dynamically adds the given machines to an AppScale deployment, with the
-    specified roles.
+    """Queries the AppController for a message that the server wants to send
+    to the tools.
 
-    Args:
-      clean: A boolean indicating whether the clean parameter should be
-        passed to terminate.rb.
     Returns:
-      The result of executing the SOAP call on the remote AppController.
+      The message from the AppController in JSON with format :
+      {'ip':ip, 'status': status, 'output':output}
     """
     return self.run_with_timeout(self.DEFAULT_TIMEOUT * 5, "Error",
                                  self.DEFAULT_NUM_RETRIES,
--- a/appscale/tools/appcontroller_client.py
+++ b/appscale/tools/appcontroller_client.py
@@ -227,6 +227,44 @@ class AppControllerClient():
     return self.run_with_timeout(self.DEFAULT_TIMEOUT, "Error", self.DEFAULT_NUM_RETRIES,
       self.server.start_roles_on_nodes, roles_to_nodes, self.secret)
 
+  def is_appscale_terminated(self):
+    """Dynamically adds the given machines to an AppScale deployment, with the
+    specified roles.
+
+    Returns:
+      A boolean indicating whether appscale has finished running terminate
+        on all nodes.
+    """
+    return self.run_with_timeout(self.DEFAULT_TIMEOUT, "Error",
+                                 self.DEFAULT_NUM_RETRIES,
+                                 self.server.is_appscale_terminated, self.secret)
+
+  def run_terminate(self, clean):
+    """Dynamically adds the given machines to an AppScale deployment, with the
+    specified roles.
+
+    Args:
+      clean: A boolean indicating whether the clean parameter should be
+        passed to terminate.rb.
+    Returns:
+      The result of executing the SOAP call on the remote AppController.
+    """
+    return self.server.run_terminate(clean, self.secret)
+
+
+  def receive_server_message(self, message_sender):
+    """Dynamically adds the given machines to an AppScale deployment, with the
+    specified roles.
+
+    Args:
+      clean: A boolean indicating whether the clean parameter should be
+        passed to terminate.rb.
+    Returns:
+      The result of executing the SOAP call on the remote AppController.
+    """
+    return self.run_with_timeout(self.DEFAULT_TIMEOUT * 2, "Error",
+                                 self.DEFAULT_NUM_RETRIES,
+                                 self.server.receive_server_message, message_sender, self.secret)
 
   def stop_app(self, app_id):
     """Tells the AppController to no longer host the named application.
--- a/appscale/tools/appcontroller_client.py
+++ b/appscale/tools/appcontroller_client.py
@@ -234,10 +234,10 @@ class AppControllerClient():
       A boolean indicating whether appscale has finished running terminate
         on all nodes.
     """
-    return self.run_with_timeout(self.DEFAULT_TIMEOUT,
-                                 "Error: Client Timed Out",
+    return self.run_with_timeout(self.DEFAULT_TIMEOUT, "Error",
                                  self.DEFAULT_NUM_RETRIES,
-                                 self.server.is_appscale_terminated, self.secret)
+                                 self.server.is_appscale_terminated,
+                                 self.secret)
 
   def run_terminate(self, clean):
     """Tells the AppController to terminate AppScale on the deployment.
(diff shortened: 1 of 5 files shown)


appscale/tools/remote_helper.py
```diff
@@ -938,41 +940,45 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
...
+                                          args=(clean,))
+      terminate_thread.start()
+      terminated_successfully = True
+      log_dump = ""
+      while not acc.is_appscale_terminated():
+        # for terminate receive_server_message will return a JSON string that
+        #  is a list of dicts with keys: ip, status, output
+        output_list = yaml.safe_load(acc.receive_server_message())
+        if output_list == "Error":
+          time.sleep(0.3)
```
obino: Should we have a constant for such a small wait time?

whoarethebritons: You know I think this wait time might not even be necessary anymore... I'll look into it.

whoarethebritons: it is definitely not needed, I will remove it.

appscale/tools/appcontroller_client.py
```diff
@@ -227,6 +227,46 @@ def start_roles_on_nodes(self, roles_to_nodes):
     return self.run_with_timeout(self.DEFAULT_TIMEOUT, "Error", self.DEFAULT_NUM_RETRIES,
       self.server.start_roles_on_nodes, roles_to_nodes, self.secret)
 
+  def is_appscale_terminated(self):
+    """Dynamically adds the given machines to an AppScale deployment, with the
```
tmarballi: Please update the doc string for this method and the two methods below.

appscale/tools/appcontroller_client.py
```diff
@@ -227,6 +227,46 @@ def start_roles_on_nodes(self, roles_to_nodes):
...
+    Args:
+      clean: A boolean indicating whether the clean parameter should be
+        passed to terminate.rb.
+    Returns:
+      The result of executing the SOAP call on the remote AppController.
+    """
+    return self.run_with_timeout(self.DEFAULT_TIMEOUT * 5, "Error",
+                                 self.DEFAULT_NUM_RETRIES,
+                                 self.server.receive_server_message,
+                                 self.DEFAULT_TIMEOUT * 4,
```
tmarballi: You might have told me this before, but can you explain why do you need this second timeout in AC method here apart from the one being used by the run_with_timeout.

whoarethebritons: The server side needs to also have a timeout so that it doesn't clear the queue of messages because it thinks the client got them. I made the server timeout smaller than the client timeout so that there wouldn't be any issues with it. Without the timeout on the server side the client would run the method and then time out possibly before the server returns a message, however the server won't know the client timed out and will assume the message is received and clear the message.

tmarballi: @whoarethebritons Ok got it! It makes sense to me now, Thanks 👍

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2641/

appscale/tools/remote_helper.py
```diff
@@ -940,41 +942,46 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
 
     acc = AppControllerClient(shadow_host, secret)
     try:
-      all_ips = acc.get_all_public_ips()
+      terminate_thread = threading.Thread(target=acc.run_terminate,
+                                          args=(clean,))
+      terminate_thread.start()
+      terminated_successfully = True
+      log_dump = ""
+      while not acc.is_appscale_terminated():
```
cdonati: `run_with_timeout` will return the default string passed in upon timeout. Any non-empty string will evaluate as true here, which I'm not sure you want.

appscale/tools/appcontroller_client.py
```diff
@@ -227,6 +227,42 @@ def start_roles_on_nodes(self, roles_to_nodes):
...
+
+    Args:
+      clean: A boolean indicating whether the clean parameter should be
+        passed to terminate.rb.
+    Returns:
+      The result of executing the SOAP call on the remote AppController.
+    """
+    return self.server.run_terminate(clean, self.secret)
+
+  def receive_server_message(self):
```
cdonati: I'm not sure this pattern is a good idea. Imagine a case where the request gets through to the server, the server clears the messages in the queue, but a connection issue causes a timeout on the client side. The message would essentially be lost, and the client will keep checking for the missing message.

My suggestion would be for `run_terminate` to return a unique string that associates messages with a given request. You can then add a function to get the status of a particular request instead of a generic function to get all waiting messages.

appscale/tools/remote_helper.py
```diff
@@ -940,41 +942,46 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
...
+        # for terminate receive_server_message will return a JSON string that
+        #  is a list of dicts with keys: ip, status, output
+        output_list = yaml.safe_load(acc.receive_server_message())
+        if hasattr(output_list, "startswith") and \
```
cdonati: Checking your data structures should probably be done at a lower level. In this case, I would recommend raising an exception in `receive_server_message` if there is an issue.

appscale/tools/remote_helper.py
```diff
@@ -940,41 +942,46 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
 
     acc = AppControllerClient(shadow_host, secret)
     try:
-      all_ips = acc.get_all_public_ips()
+      terminate_thread = threading.Thread(target=acc.run_terminate,
```
cdonati: In the interest in making things simpler for the client, I would recommend making this asynchronous on the server side since you are already polling on the client side. In other words, the server function should return immediately (doing the work in a separate thread), and the client should not need to use threads.

appscale/tools/remote_helper.py
```diff
@@ -940,41 +942,46 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
...
+          log_dump += output_list
+          continue
+        for node in output_list:
+          if node.get("status"):
+            AppScaleLogger.success("Node at {node_ip}: {status}".format(
+              node_ip=node.get("ip"), status="Terminated Successfully"))
+          else:
+            AppScaleLogger.warn("Node at {node_ip}: {status}".format(
+              node_ip=node.get("ip"), status="Did not terminate successfully"))
+            terminated_successfully &= False
```
cdonati: I'm not sure what this operator does to a boolean, but is there a reason this can't be a simple `=`?

appscale/tools/remote_helper.py
```diff
@@ -940,41 +942,46 @@ def terminate_virtualized_cluster(cls, keyname, is_verbose):
...
+                                          output=node.get("output"))
+          AppScaleLogger.verbose("Output of node at {node_ip}:\n"\
+                                 "{output}".format(node_ip=node.get("ip"),
+                                                   output=node.get("output")),
+                                 is_verbose)
+      terminate_thread.join()
+      if not terminated_successfully:
+        AppScaleLogger.warn("Some nodes failed terminating")
+        LocalState.generate_crash_log(AppControllerException, log_dump)
+      cls.stop_remote_appcontroller(shadow_host, keyname, is_verbose)
```
cdonati: It doesn't look like `generate_crash_log` stops execution. Do we want to stop the controller on the shadow if some nodes failed terminating?

appscale/tools/appcontroller_client.py
whoarethebritons: @cdonati I agree with your suggestion, I had written the methods to be generic so I could later add the request id principle onto them.

However, the client will not keep checking for the missing message because the client doesn't "think" of how many messages it is meant to receive so it will not keep checking. However it will miss the message completely if that were to happen.

appscale/tools/remote_helper.py
whoarethebritons: @cdonati The machines that would be marked as not terminated successfully would be the machines whose messages we got back but had false associated with them. Because the server will have tried to run terminate on every node even if we get to here. The tools side is the only place where the head node gets terminated so I thought that if one or two nodes did not terminate successfully the head node should still come down and just let the user know what failed to come down. If all the nodes fail to come down there is something bigger wrong and we would not be able to run terminate the next time either.

appscale/tools/appcontroller_client.py
cdonati: I mistakenly thought that the client was making a decision based on the response from this call. However, I see now that a lost message would just affect the log output. It's okay with me if you want until another pr to change the message passing.

appscale/tools/remote_helper.py
cdonati: Unfortunately, that would mean that there is no easy way for the user to retry the terminate, and leaving processes running on other nodes leads to pretty nasty side effects on subsequent ups.

> If all the nodes fail to come down there is something bigger wrong and we would not be able to run terminate the next time either.

Why wouldn't we be able to run a terminate again if the shadow's controller is still running?

appscale/tools/appcontroller_client.py
cdonati: Sorry, after re-reading the method, it does look like `terminated_successfully` is set based on `receive_server_message`, so there is still the potential for a terminate to fail on some nodes without the user knowing about it.

appscale/tools/remote_helper.py
whoarethebritons: @cdonati I'm fine with moving the head node's shut down, I would like to make sure that everything needed to run a down would still be present on the system though so that they could try again, currently I do not think that is the case.

appscale/tools/appcontroller_client.py
whoarethebritons: @cdonati what if at the end I counted how many nodes were supposed to have terminated and compared?

cdonati: I think that would be fine for now. We may need to revisit when we look at autoscaling if there are machines that are part of the deployment that the tools don't know about.

appscale/tools/remote_helper.py
cdonati: I don't think you need to move it. You can just raise an exception after `generate_crash_log`.

The terminate script that the controller executes over ssh should be idempotent, and as far as I know, it doesn't require anything else to be present on the system.

appscale/tools/appcontroller_client.py
whoarethebritons: @cdonati Ok yea, we could have them call the AC on the head node to get how many nodes there should be when that time comes. Working on fixing this right now.

appscale/tools/remote_helper.py
whoarethebritons: @cdonati What if I add an additional check for if the user specified terminate? If the user wanted to terminate the instances they might not want it to stop for an unreachable node.

cdonati: I'm okay with that for now unless anyone has objections. We'll have to revisit if we ever decide to support `terminate` with `clean` since that could leave volume data behind.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2706/
https://ocd.appscale.com:8080/job/Tools-test/1447/

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2720/

PR #582, opened 2017-02-07
Fix verbose and listing scale set instances during terminate

tmarballi:


tmarballi:
Will make separate PRs for different issues.

PR #585, opened 2017-02-08
Fix the listing of scale set instances names

tmarballi:
This PR addresses that the VMSS vm list method cannot be used directly as a list and it needs to be looped into to extract the instance name.

### DIFF
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -567,9 +567,10 @@ class AzureAgent(BaseAgent):
     vmss_delete_threads = []
     deleted_instance_ids = []
     for vmss in vmss_list:
-      deleted_instance_ids.extend(
-        compute_client.virtual_machine_scale_set_vms.list(resource_group,
-                                                          vmss.name))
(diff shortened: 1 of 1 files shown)


appscale/tools/agents/azure_agent.py
```diff
@@ -567,31 +567,37 @@ def terminate_instances(self, parameters):
     vmss_delete_threads = []
     deleted_instance_ids = []
     for vmss in vmss_list:
-      deleted_instance_ids.extend(
-        compute_client.virtual_machine_scale_set_vms.list(resource_group,
-                                                          vmss.name))
+      vm_list = compute_client.virtual_machine_scale_set_vms.list(
+        resource_group, vmss.name)
+      for vm in vm_list:
+        deleted_instance_ids.append(vm.name)
```
menivaitsi: Maybe I'm confused but are there instance ids already deleted here?
If not, please rename to _instance_ids_to_delete_.

tmarballi: No you are right, it is misleading. Will rename the variable.

PR #586, opened 2017-02-08
Cleanup keyname after terminating azure instances

tmarballi:


appscale/tools/agents/azure_agent.py
```diff
@@ -723,28 +723,41 @@ def cleanup_state(self, parameters):
...
     for network in virtual_networks:
       result = network_client.virtual_networks.delete(resource_group, network.name)
       resource_name = 'Virtual Network' + ':' + network.name
       self.sleep_until_delete_operation_done(result, resource_name,
                                              self.MAX_SLEEP_TIME, verbose)
+      AppScaleLogger.verbose("Virtual Network {} has been successfully deleted.".
+                             format(network.name), verbose)
+
+    AppScaleLogger.log("Virtual Network(s) have been successfully deleted.")
+    LocalState.cleanup_keyname(parameters[self.PARAM_KEYNAME])
```
obino: Why do we delete the keys here? Shouldn't be the tools (not the agent) to take care of it? If I read correctly the other agents cleanup state in the cloud not on the file system.

menivaitsi: @obino @tmarballi It used to be in the proper place IMHO: https://github.com/AppScale/appscale-tools/blob/b9c7fd660fc12de575a3af49636ef3496b8dc46a/appscale/tools/appscale_tools.py#L627

but it was removed from there.

tmarballi: So there is a slightly different behavior in two sets of agents, EC2/Euca and GCE/Azure. For EC2 agent we create a keypair in the cloud itself, so in the clean up stage, we use the delete keypair which cleans up state in the cloud. But for GCE & Azure, we create the keys on the local file system, so we need a way for both of them to clean up state on terminate, otherwise we would not be able to reuse the keyname.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/2714/

appscale/tools/agents/azure_agent.py
tmarballi: @menivaitsi Sorry I somehow saw your comment after I posted mine. You are right about the clean up files being there, and I believe it was removed when we separated the regular down and --terminate. The cleanup_appscale_files method here https://github.com/AppScale/appscale-tools/blob/master/appscale/tools/local_state.py#L858 just removes the json locations and the secret. But the check before starting with the same keyname is for the ssh keys and the keyname file, which when present we would not be able to do an up. Now the method I added, would remove the rest of the files which just the 'down' does not clean.

PR #593, opened 2017-02-16
catch BadSecretException when uploading app

whoarethebritons:
creates BadSecretException which is a child of AppControllerException
stops potential infinite loop

https://ocd.appscale.com:8080/job/Daily%20Build/2784/

### DIFF
--- a/appscale/tools/appcontroller_client.py
+++ b/appscale/tools/appcontroller_client.py
@@ -16,8 +16,9 @@ import SOAPpy
 # AppScale-specific imports
 from appscale_logger import AppScaleLogger
 from custom_exceptions import AppControllerException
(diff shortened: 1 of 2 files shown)


appscale/tools/appcontroller_client.py
```diff
@@ -471,6 +472,10 @@ def does_user_exist(self, username, silent=False):
           return False
         else:
           raise Exception(user_exists)
+      except BadSecretException as exception:
+        AppScaleLogger.warn("Exception when checking if a user exists: {0}".
+                            format(exception))
+        break
```
cdonati: Returning `None` here seems unusual to me. Is there a reason for this rather than raising an exception or logging and stopping?

PR #594, opened 2017-02-16
ssh to roles and indices

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2784/

### DIFF
--- a/appscale/tools/appscale.py
+++ b/appscale/tools/appscale.py
@@ -403,25 +403,30 @@ Available commands:
     contents = self.read_appscalefile()
     contents_as_yaml = yaml.safe_load(contents)
 
-    # make sure the user gave us an int for node
-    try:
-      index = int(node)
-    except ValueError:
-      raise TypeError("Usage: appscale ssh <node id to ssh to>")
-
     if 'keyname' in contents_as_yaml:
       keyname = contents_as_yaml['keyname']
     else:
       keyname = "appscale"
 
-    nodes = self.get_nodes(keyname)
-    # make sure there is a node at position 'index'
+    if node is None:
+      node = "shadow"
+
     try:
+      index = int(node)
+      nodes = self.get_nodes(keyname)
+      # make sure there is a node at position 'index'
       ip = nodes[index]['public_ip']
     except IndexError:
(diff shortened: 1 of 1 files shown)


appscale/tools/appscale.py
```diff
@@ -403,25 +403,30 @@ def ssh(self, node):
...
     try:
+      index = int(node)
+      nodes = self.get_nodes(keyname)
+      # make sure there is a node at position 'index'
       ip = nodes[index]['public_ip']
     except IndexError:
       raise AppScaleException("Cannot ssh to node at index " +
-        str(index) + ", as there are only " + str(len(nodes)) +
-        " in the currently running AppScale deployment.")
+                              ", as there are only " + str(len(nodes)) +
```
tmarballi: What do you think will happen in case of non load balancers which would be only returning private ips?  Can we put a check for that?

whoarethebritons: @tmarballi we would get "ssh: Could not resolve hostname : Name or service not known". I think most people would be just using "appscale ssh" to get to the head node... maybe I could put a help text that would explain like: "use an index or a role in [list of roles], if you have a deployment with private ips this may not work". However currently the appscale ssh [index] doesn't work if there are private ips.

appscale/tools/appscale.py
```diff
@@ -403,25 +403,30 @@ def ssh(self, node):
     contents = self.read_appscalefile()
     contents_as_yaml = yaml.safe_load(contents)
 
-    # make sure the user gave us an int for node
-    try:
-      index = int(node)
```
menivaitsi: Are there use cases that'd we'd wanna keep this block of code for?

whoarethebritons: @menivaitsi just to not deprecate it I guess.

appscale/tools/appscale.py
tmarballi: Yeah makes sense @whoarethebritons, I would definitely vote for putting in some help text so that ssh error would not throw someone off. Or maybe catch that exception, and have a helpful log statement.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2827/

PR #596, opened 2017-02-25
Remove unnecessary sleep for cluster starts

cdonati:


appscale/tools/remote_helper.py
```diff
@@ -254,7 +254,7 @@ def start_head_node(cls, options, my_id, node_layout):
       if agent.PARAM_REGION in params:
         additional_params[agent.PARAM_REGION] = params[agent.PARAM_REGION]
 
-    time.sleep(10)  # gives machines in cloud extra time to boot up
+      time.sleep(10)  # gives machines in cloud extra time to boot up
```
obino: Is this sleep actually necessary? And if so, do we know why 10 seconds is enough?

cdonati: I doubt the sleep is necessary, but removing it for all cloud deployments would require a lot of testing.

menivaitsi: I believe this was originally placed for GCE but was moved around and ended up here.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2808/

PR #598, opened 2017-03-03
mark nodes besides load_balancers as open during a clean

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2840/

### DIFF
--- a/appscale/tools/local_state.py
+++ b/appscale/tools/local_state.py
@@ -420,6 +420,29 @@ class LocalState(object):
 
 
   @classmethod
+  def clean_local_metadata(cls, keyname):
+    try:
+      with open(cls.get_locations_json_location(keyname), 'r') as file_handle:
+        file_contents = yaml.safe_load(file_handle.read())
+        if isinstance(file_contents, list):
+          cls.upgrade_json_file(keyname)
+          file_handle.seek(0)
+          file_contents = json.loads(file_handle.read())
+        cleaned_nodes = []
+        for node in file_contents.get('node_info'):
+          if 'load_balancer' not in node.get('jobs'):
+            node['jobs'] = ['open']
+          cleaned_nodes.append(node)
+        file_contents['node_info'] = cleaned_nodes
+        # and now we can write the json metadata file
+      with open(cls.get_locations_json_location(keyname), 'w') as file_handle:
+        file_handle.write(json.dumps(file_contents))
+    except IOError:
+      raise BadConfigurationException("Couldn't read from locations file, "
+                                      "AppScale may not be running with "
+                                      "keyname {0}".format(keyname))
+
+  @classmethod
   def get_infrastructure_option(cls, tag, keyname):
     """Reads the JSON-encoded metadata on disk and returns the value for
     the key 'tag' from the dictionary retrieved using the key
--- a/appscale/tools/local_state.py
+++ b/appscale/tools/local_state.py
@@ -421,9 +421,20 @@ class LocalState(object):
 
   @classmethod
   def clean_local_metadata(cls, keyname):
+    """Takes the existing JSON-encoded metadata on disk and assigns all nodes
+    besides load_balancers (because of public ips) to "open".
+
+    Args:
+      keyname: A str that represents an SSH keypair name, uniquely identifying
+        this AppScale deployment.
+    Raises:
+      BadConfigurationException: If there is no JSON-encoded metadata file
+        named after the given keyname.
+    """
     try:
-      with open(cls.get_locations_json_location(keyname), 'r') as file_handle:
+      with open(cls.get_locations_json_location(keyname), 'r+') as file_handle:
         file_contents = yaml.safe_load(file_handle.read())
+        # Compatibility support for previous versions of locations file.
         if isinstance(file_contents, list):
           cls.upgrade_json_file(keyname)
           file_handle.seek(0)
@@ -434,13 +445,12 @@ class LocalState(object):
             node['jobs'] = ['open']
           cleaned_nodes.append(node)
         file_contents['node_info'] = cleaned_nodes
-        # and now we can write the json metadata file
-      with open(cls.get_locations_json_location(keyname), 'w') as file_handle:
+        # Now we write the JSON file after our changes.
+        file_handle.seek(0)
+        file_handle.truncate()
         file_handle.write(json.dumps(file_contents))
     except IOError:
-      raise BadConfigurationException("Couldn't read from locations file, "
-                                      "AppScale may not be running with "
-                                      "keyname {0}".format(keyname))
+      raise BadConfigurationException("Couldn't read from locations file.")
 
   @classmethod
   def get_infrastructure_option(cls, tag, keyname):
--- a/appscale/tools/node_layout.py
+++ b/appscale/tools/node_layout.py
@@ -740,20 +740,20 @@ class NodeLayout():
             # Locations JSON is incorrect if we get here.
             return None
           break
-    if len(nodes_copy) > 0:
+    try:
       for open_node in open_nodes:
-        for _, node in enumerate(nodes_copy):
-          # Match nodes based on jobs/roles
-          nodes_copy.remove(node)
-          roles = node.roles
-          node.from_json(open_node)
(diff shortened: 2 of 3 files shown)


appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
 
 
   @classmethod
+  def clean_local_metadata(cls, keyname):
+    try:
+      with open(cls.get_locations_json_location(keyname), 'r') as file_handle:
+        file_contents = yaml.safe_load(file_handle.read())
```
cdonati: I'm curious about this yaml load. Do we ever expect this file to contain non-json data?

appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
...
+        if isinstance(file_contents, list):
```
menivaitsi: Please add a comment explaining why we have this block.

menivaitsi: If I understand it correctly, it's for backwards compatibility.

appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
...
+          cls.upgrade_json_file(keyname)
+          file_handle.seek(0)
+          file_contents = json.loads(file_handle.read())
+        cleaned_nodes = []
+        for node in file_contents.get('node_info'):
+          if 'load_balancer' not in node.get('jobs'):
+            node['jobs'] = ['open']
+          cleaned_nodes.append(node)
+        file_contents['node_info'] = cleaned_nodes
+        # and now we can write the json metadata file
```
menivaitsi: Full sentences.

appscale/tools/node_layout.py
```diff
@@ -737,7 +740,20 @@ def from_locations_json_list(self, locations_nodes_list):
             # Locations JSON is incorrect if we get here.
             return None
           break
-
+    if len(nodes_copy) > 0:
+      for open_node in open_nodes:
+        for _, node in enumerate(nodes_copy):
+          # Match nodes based on jobs/roles
```
menivaitsi: Add period at the end.

appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
...
+      with open(cls.get_locations_json_location(keyname), 'w') as file_handle:
+        file_handle.write(json.dumps(file_contents))
```
cdonati: When you already have a file handle, I think it's clearer just to use `.load` and `.dump`.

appscale/tools/node_layout.py
```diff
@@ -737,7 +740,20 @@ def from_locations_json_list(self, locations_nodes_list):
...
+          nodes_copy.remove(node)
+          roles = node.roles
+          node.from_json(open_node)
+          node.roles = roles
+          if node.is_valid():
+            nodes.append(node)
+          else:
+            # Locations JSON is incorrect if we get here.
```
menivaitsi: Shouldn't we log it if there's a problem with the locations layout?

appscale/tools/node_layout.py
```diff
@@ -737,7 +740,20 @@ def from_locations_json_list(self, locations_nodes_list):
...
+            return None
```
tmarballi: Do we handle the incorrect location JSON where this method is called, considering we return None here?

appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
 
 
   @classmethod
+  def clean_local_metadata(cls, keyname):
+    try:
```
tmarballi: Please add a doc string here.

appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
...
+          if 'load_balancer' not in node.get('jobs'):
+            node['jobs'] = ['open']
+          cleaned_nodes.append(node)
+        file_contents['node_info'] = cleaned_nodes
+        # and now we can write the json metadata file
+      with open(cls.get_locations_json_location(keyname), 'w') as file_handle:
+        file_handle.write(json.dumps(file_contents))
+    except IOError:
+      raise BadConfigurationException("Couldn't read from locations file, "
+                                      "AppScale may not be running with "
```
cdonati: This message may be confusing since in this case it would only show up after we've stopped services. I would stick with just what we know. I think "Couldn't read from locations file." would suffice.

appscale/tools/node_layout.py
```diff
@@ -737,7 +740,20 @@ def from_locations_json_list(self, locations_nodes_list):
             # Locations JSON is incorrect if we get here.
             return None
           break
-
+    if len(nodes_copy) > 0:
+      for open_node in open_nodes:
+        for _, node in enumerate(nodes_copy):
```
cdonati: If you don't need the index, you probably don't need enumerate. Also, it looks like this is always just popping the first item from the list, so I'm not sure why iterating over `nodes_copy` is necessary.

appscale/tools/node_layout.py
whoarethebritons: @tmarballi we don't handle it because NodeLayout isn't really equipped to in my opinion, returning None makes the caller raise an exception (because exceptions are never raised in this file for some reason)

appscale/tools/node_layout.py
whoarethebritons: @menivaitsi returning None makes the caller raise an exception (because exceptions are never raised in this file for some reason) that portion is in remote_helper

appscale/tools/local_state.py
whoarethebritons: @menivaitsi yes its for backwards compatibility, I will put in the comment.

appscale/tools/node_layout.py
```diff
@@ -737,7 +740,20 @@ def from_locations_json_list(self, locations_nodes_list):
             # Locations JSON is incorrect if we get here.
             return None
           break
-
+    if len(nodes_copy) > 0:
```
cdonati: To reduce nesting, you can remove this check. If you pop from `nodes_copy`, you can catch `IndexError`.

appscale/tools/local_state.py
whoarethebritons: @cdonati yaml.safe_load just does a better job of decoding JSON into python objects, primary example is a JSON load will load strings as unicode strings and string comparisons will fail.

cdonati: String comparisons between unicode and non-unicode strings should work just fine.
```
>>> u'value' == 'value'
True
```
Standard libs should be preferred over 3rd party ones in most cases unless they make the code more difficult to maintain.

appscale/tools/local_state.py
```diff
@@ -420,6 +420,29 @@ def update_local_metadata(cls, options, db_master, head_node):
 
 
   @classmethod
+  def clean_local_metadata(cls, keyname):
```
whoarethebritons: @menivaitsi fixed but its not registering... it marked Tanvi's as fixed.

appscale/tools/node_layout.py
```diff
@@ -737,7 +740,20 @@ def from_locations_json_list(self, locations_nodes_list):
...
+        # Match nodes based on jobs/roles.
+        roles = node.roles
+        node.from_json(open_node)
+        node.roles = roles
+        if node.is_valid():
+          nodes.append(node)
+        else:
+          # Locations JSON is incorrect if we get here.
+          return None
+    except IndexError:
```
cdonati: Only the pop needs the try/except.

appscale/tools/node_layout.py
tmarballi: @whoarethebritons Yeah makes sense! I think the None that is returned, is checked for in the remote helper, which throws an exception accordingly.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2885/

PR #599, opened 2017-03-06
Update status command

jeanleonov:
Here is updated status command for tools. It works with updated AppController, which is not ready to be merged ([appscale/cluster-stats-refactoring](https://github.com/AppScale/appscale/compare/cluster-stats-refactoring)).
Tomorrow I'll do minor change requested by Meni.
If you're interested you can take a look and leave some comments.

Should be merged only together with [appscale/cluster-stats-refactoring](https://github.com/AppScale/appscale/compare/cluster-stats-refactoring)

p.s.: sorry that I created branches directly in appscale and appscale-tools repositories. I'll do branching in the proper way for future work.

### DIFF
--- a/appscale/tools/appscale_tools.py
+++ b/appscale/tools/appscale_tools.py
@@ -225,9 +225,8 @@ class AppScaleTools(object):
     login_host = LocalState.get_login_host(options.keyname)
     login_acc = AppControllerClient(login_host,
       LocalState.get_secret_key(options.keyname))
-    import pprint
     stats = login_acc.get_cluster_stats()
-    AppScaleLogger.log(pprint.pprint(stats))
+    AppScaleLogger.log(str(stats))
     AppScaleLogger.success("View status information about your AppScale " + \
       "deployment at http://{0}:{1}/status".format(login_host,
       RemoteHelper.APP_DASHBOARD_PORT))
@@ -394,18 +393,17 @@ class AppScaleTools(object):
     # through them for the http port the app can be reached on.
     http_port = None
     for _ in range(cls.MAX_RETRIES + 1):
-      result = acc.get_all_stats()
+      result = acc.get_app_info_map()
       try:
-        json_result = json.loads(result)
-        apps_result = json_result['apps']
-        current_app = apps_result[options.appname]
-        http_port = current_app['http']
+        current_app = result[options.appname]
+        http_port = current_app['nginx']
         if http_port:
           break
         time.sleep(cls.SLEEP_TIME)
       except (KeyError, ValueError):
-        AppScaleLogger.verbose("Got json error from get_all_data result.",
-            options.verbose)
+        AppScaleLogger.verbose(
+          "Got json error from get_app_info_map result:\n{}".format(result),
+          options.verbose)
         time.sleep(cls.SLEEP_TIME)
     if not http_port:
       raise AppScaleException(
@@ -735,18 +733,17 @@ class AppScaleTools(object):
     # through them for the http port the app can be reached on.
     http_port = None
     for _ in range(cls.MAX_RETRIES + 1):
-      result = acc.get_all_stats()
+      result = acc.get_app_info_map()
       try:
-        json_result = json.loads(result)
-        apps_result = json_result['apps']
-        current_app = apps_result[app_id]
-        http_port = current_app['http']
+        current_app = result[app_id]
+        http_port = current_app['nginx']
         if http_port:
           break
         time.sleep(cls.SLEEP_TIME)
       except (KeyError, ValueError):
-        AppScaleLogger.verbose("Got json error from get_all_data result.",
-            options.verbose)
+        AppScaleLogger.verbose(
+          "Got json error from get_app_info_map result:\n{}".format(result),
+          options.verbose)
         time.sleep(cls.SLEEP_TIME)
     if not http_port:
       raise AppScaleException(
--- a/appscale/tools/appscale_tools.py
+++ b/appscale/tools/appscale_tools.py
@@ -261,9 +261,9 @@ class AppScaleTools(object):
     else:
       AppScaleLogger.log("-"*76)
 
-    cls._print_cluster_summary(nodes, apps)
+    cls._print_cluster_summary(nodes, invisible_nodes, apps)
     cls._print_apps(apps)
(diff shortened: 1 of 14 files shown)


jeanleonov:
Wow=) Great static analizer

menivaitsi:
@jeanleonov great catch indeed!

appscale/tools/appscale_tools.py
```diff
@@ -51,6 +57,12 @@ def async_layout_upgrade(ip, keyname, script, error_bucket, verbose=False):
     error_bucket.put(ssh_error)
 
 
+MIN_FREE_DISK_DB = 40.0
```
cdonati: I would recommend putting these thresholds on the server side so that we can present consistent warnings on both the dashboard and the tools. If we ever decide to change the value, it will be easier if it's just defined in one spot. We may also want to allow the user to configure these thresholds eventually.

I'm not sure about the best way to implement it, though. One idea would be to include a "warnings" section in the stats passed by the controller. Unfortunately, those aren't technically stats, so I'm not sure if that's the cleanest way.

Another possibility would be to add a new controller endpoint to retrieve current thresholds. However, that has the downside of requiring fixed categories. For example, if we wanted to add something like `MAX_IOPS` (just as a contrived example), we would need to change to tools so that they could make sense of the new category.

Do you have any ideas about how those warnings should be communicated?

setup.py
```diff
@@ -58,7 +58,8 @@
     'setuptools>=11.3,<34',
     'SOAPpy',
     'termcolor',
-    'wstools==0.4.3'
+    'wstools==0.4.3',
+    'tabulate==0.7.7'
```
cdonati: Cool library. 👍

appscale/tools/appscale_tools.py
obino: I agree with @cdonati: the server side should returns those warning. I would be even more generic: the server site should return warnings (as in error codes/strings) not even the parameters to check for errors. The tools should just report whatever warning it reads. This would get a long way once we get hermes fully implemented :)

appscale/tools/appscale_tools.py
```diff
@@ -215,29 +227,193 @@ def add_keypair(cls, options):
...
+        LocalState.get_secret_key(options.keyname))
+      all_public_ips = login_acc.get_all_public_ips()
+      cluster_stats = login_acc.get_cluster_stats()
+    except (faultType, AppControllerException, BadConfigurationException):
+      AppScaleLogger.warn("AppScale deployment is probably down")
+      raise
+
+    # Convert cluster stats to useful structures
+    node_stats = {
+      ip: next((n for n in cluster_stats if n["public_ip"] == ip), None)
```
obino: @jeanleonov not sure if I'm reading this correctly, but are we skipping all nodes that havea  a public IP here?

appscale/tools/appscale_tools.py
```diff
@@ -215,29 +227,193 @@ def add_keypair(cls, options):
 
   @classmethod
   def describe_instances(cls, options):
```
obino: this may be misleading as a name, since we have a bunch of describe_instances in the agents,,,

appscale/tools/appscale_tools.py
```diff
@@ -215,29 +227,193 @@ def add_keypair(cls, options):
...
+      (app for app in apps if app.http == RemoteHelper.APP_DASHBOARD_PORT), None
+    )
+    if dashboard and dashboard.appservers > 1:
+      AppScaleLogger.success(
+        "\nView more about your AppScale deployment at http://{}:{}/status"
+        .format(login_host, RemoteHelper.APP_DASHBOARD_PORT)
+      )
+    else:
+      AppScaleLogger.log(
+        "\nAs soon as AppScale Dashboard is started you can visit it at "
```
obino: I think at the time when one can execute status, the dashbiard should be already up, so the message should be more simple, something like: You can visit the AppScale Dsahboard at ..."

jeanleonov: Not necessarily. Sometimes I give up waiting for dashboard and just `Ctrl+C` `appscale up` command, but after that I still want to see trues status message. Additionally, sometimes AppController just can't start Dashboard at all. For example if there is lack of memory, or there is some bug=)

appscale/tools/appscale_tools.py
jeanleonov: It is not the most well readable construction, but I found that many experienced python devs use it and it seems to be like well known way to get the first match in list.
Function `next` returns the next element for iterator. In this case I pass iterator is generator `n for n in cluster_stats if n["public_ip"] == ip`. This generator iterates though list of nodes and yield only those which match public IP.

appscale/tools/appscale_tools.py
jeanleonov: @cdonati you're right. But maybe it makes sense to refactor AppController first? It doesn't look like a critical functionality, so imho it can wait for REST-like architecture of AppController. If I don't move warning-functionality to AppController, do you want me to remove it from tools?

appscale/tools/local_state.py
```diff
@@ -104,7 +104,7 @@ def ensure_appscale_isnt_running(cls, keyname, force):
 
       acc = AppControllerClient(login_host, secret_key)
       try:
-        acc.get_status()
+        acc.get_all_public_ips()
```
jeanleonov: Is used just to check accessibility of login node.

obino: Got it, makes sense. If you could pick something more innocuos would be best (I think this function should be removed to favored the nodes layout we have for example), maybe is_done_loading, or or get_role_info etc...

appscale/tools/appscale_tools.py
obino: Well that's a Ctrl-C though, which means you won't reach this message, won't you?

jeanleonov: We can reach this message. e.g.: I gave up waiting for appscale to start dashboard but after 5min working on other stuff I decided  to check if autoscaler found a resources for AppServer for dashboard. If dashboard is still not started I'll get this message.

jeanleonov:
[Build #2966](https://ocd.appscale.com:8080/job/Daily%20Build/2966/) - single failure in okay tests was manually verified and is irrelevant to this PR

PR #600, opened 2017-03-07
Use AppScalefile as fallback source for IPs

cdonati:
After an upgrade from <= 3.1.0, the user may not have a locations JSON file. This change allows the tools to still check if the ssh key is valid by using the AppScalefile to generate a list of all addresses in a cluster.

### DIFF
--- a/appscale/tools/appscale.py
+++ b/appscale/tools/appscale.py
@@ -337,7 +337,11 @@ Available commands:
     if not os.path.exists(ssh_key_location):
       return False
 
-    all_ips = LocalState.get_all_public_ips(keyname)
+    try:
+      all_ips = LocalState.get_all_public_ips(keyname)
+    except BadConfigurationException:
+      # If this is an upgrade from 3.1.0, there may not be a locations JSON.
+      all_ips = set(run_instances_opts.ips.values())
 
     # If a login node is defined, use that to communicate with other nodes.
     node_layout = NodeLayout(run_instances_opts)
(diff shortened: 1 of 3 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2866/

appscale/tools/appscale.py
```diff
@@ -337,7 +337,11 @@ def valid_ssh_key(self, config, run_instances_opts):
     if not os.path.exists(ssh_key_location):
       return False
 
-    all_ips = LocalState.get_all_public_ips(keyname)
+    try:
+      all_ips = LocalState.get_all_public_ips(keyname)
+    except BadConfigurationException:
+      # If this is an upgrade from 3.1.0, there may not be a locations JSON.
+      all_ips = set(run_instances_opts.ips.values())
```
obino: Few lines below there is a look and then an SSH connection to all IPs, What happens if these IPs are say wrong (like node-2, node-3...)?

cdonati: This function is only run when 'infrastructure' is not in the AppScalefile. I'm open to suggestions for making it more defensive, though.

whoarethebritons: Would checking to see if "node" is in the ip strings work as a first check? However, what would we do if that was the case...

cdonati: Yes, I can add an assert to make sure that we are using IP addresses. I would guess that we want to fail if we are not.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/2994/

PR #601, opened 2017-03-07
Allow server side to use the tools agents for autoscaling

tmarballi:
Requires appscale PR: https://github.com/AppScale/appscale/pull/2368

### DIFF
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -474,7 +474,7 @@ class AzureAgent(BaseAgent):
       vm_list = compute_client.virtual_machine_scale_set_vms.list(
         resource_group, vmss.name)
       ss_instance_count = 0
-      for vm in vm_list:
+      for _ in vm_list:
         ss_instance_count += 1
 
       if ss_instance_count >= self.MAX_VMSS_CAPACITY:
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -457,7 +457,7 @@ class AzureAgent(BaseAgent):
     adds instances (created as a part of autoscaling) to the ones which have
     additional capacity.
     Args:
-        count: The number of virtual machines to be created in the scale set.
+        count: The number of instances to be created for autoscaling.
         parameters: A dict, containing all the parameters necessary to
           authenticate this user with Azure.
     Returns:
@@ -476,8 +476,6 @@ class AzureAgent(BaseAgent):
       ss_instances = []
       for vm in vm_list:
         ss_instances.append(vm.name)
-      ss_instance_count = len(ss_instances)
-      capacity = self.MAX_VMSS_CAPACITY - ss_instance_count
 
       if len(ss_instances) >= self.MAX_VMSS_CAPACITY:
         continue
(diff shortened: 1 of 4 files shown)


tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/2874/

appscale/tools/agents/azure_agent.py
```diff
@@ -427,18 +434,89 @@ def create_linux_configuration(self, parameters):
...
+      capacity = self.MAX_VMSS_CAPACITY - ss_instance_count
+
+      scaleset = compute_client.virtual_machine_scale_sets.get(
+        resource_group, vmss.name)
+      ss_upgrade_policy = scaleset.upgrade_policy
+      ss_location = scaleset.location
+      ss_profile = scaleset.virtual_machine_profile
+      ss_overprovision = scaleset.over_provision
+
+      if capacity:
```
cdonati: It looks like you can simplify things by replacing this if/else with the following:
```
if len(ss_instances) >= self.MAX_VMSS_CAPACITY:
  continue

new_capacity = min(len(ss_instances) + count, self.MAX_VMSS_CAPACITY)
sku = ComputeSku(...capacity=new_capacity)
scaleset = ...
response = ...
self.wait_for_ss_update(new_capacity, response, vmss.name)

newly_added = new_capacity - len(ss_instances)
num_instances_added += newly_added
if num_instances_added >= count:
  break
```

appscale/tools/agents/azure_agent.py
```diff
@@ -454,6 +532,19 @@ def create_or_update_vm_scale_sets(self, count, parameters, subnet):
...
+    # While autoscaling, look through existing scale sets to check if they have
+    # capacity to hold more vms. If they do, update the scale sets with additional
+    # vms. If not, then create a new scale set for them.
+    is_autoscale = parameters['autoscale_agent']
+    num_instances_to_add = count
+    if is_autoscale in ['True', True]:
+      instances_added = self.add_instances_to_existing_ss(
+        num_instances_to_add, parameters)
+      if num_instances_to_add > instances_added:
+        count = count - instances_added
```
cdonati: I would recommend modifying `num_instances_to_add` instead of `count` just in case someone later decides to edit this function and assumes (from the docstring) that count always refers to the number the client asked for.

tmarballi: Yeah thats right, It was easier to let it modify the count because I wanted it to default to the regular logic of creating new scale sets once its exhausted the capacity of existing ones. But it makes sense to not to change the original count. I will fix that!

appscale/tools/agents/azure_agent.py
```diff
@@ -206,11 +206,18 @@ def configure_instance_security(self, parameters):
       AgentRuntimeException: If security features could not be successfully
         configured in the underlying cloud.
     """
+    is_autoscale = parameters['autoscale_agent']
+
+    # While creating instances during autoscaling, we do not need to create a
+    # new keypair or a resource group. We just make use of the existing one.
+    if is_autoscale in ['True', True]:
```
tmarballi: No, this gets set as a boolean in the Infrastructure manager client, which then gets passed around to the agent as an uppercase 'True' as a string. 

I ran into this issue when I was testing autoscaling using the tools agents, the parameters are passed around the server side as a string, but on the tools side they are passed as a boolean. Hence the check for both. Its passed from the Infrastructure Manager client to the manager via a SOAP call, so I believe that can be done only as a string.

appscale/tools/agents/azure_agent.py
```diff
@@ -427,18 +434,89 @@ def create_linux_configuration(self, parameters):
...
+        continue
+
+      scaleset = compute_client.virtual_machine_scale_sets.get(
+        resource_group, vmss.name)
+      ss_upgrade_policy = scaleset.upgrade_policy
+      ss_location = scaleset.location
+      ss_profile = scaleset.virtual_machine_profile
+      ss_overprovision = scaleset.over_provision
+
+      if count > capacity:
```
cdonati: What I meant by my previous comment was that you don't need the duplicate code here. You can replace this if/else with just the following block:
```
new_capacity = min(len(ss_instances) + count, self.MAX_VMSS_CAPACITY)
sku = ComputeSku(...capacity=new_capacity)
scaleset = ...
response = ...
self.wait_for_ss_update(new_capacity, response, vmss.name)

newly_added = new_capacity - len(ss_instances)
num_instances_added += newly_added
if num_instances_added >= count:
  break
```

appscale/tools/agents/azure_agent.py
```diff
@@ -246,7 +253,7 @@ def describe_instances(self, parameters, pending=False):
       instance_ids: A list of unique Azure VM names.
     """
     credentials = self.open_connection(parameters)
-    subscription_id = parameters[self.PARAM_SUBSCRIBER_ID]
+    subscription_id = str(parameters[self.PARAM_SUBSCRIBER_ID])
```
whoarethebritons: Is this ever not a string? (and same for all the ones below)

appscale/tools/agents/azure_agent.py
```diff
@@ -427,18 +434,82 @@ def create_linux_configuration(self, parameters):
...
+    resource_group = parameters[self.PARAM_RESOURCE_GROUP]
+    compute_client = ComputeManagementClient(credentials, subscription_id)
+
+    num_instances_added = 0
+    vmss_list = compute_client.virtual_machine_scale_sets.list(resource_group)
+    for vmss in vmss_list:
+      vm_list = compute_client.virtual_machine_scale_set_vms.list(
+        resource_group, vmss.name)
+      ss_instances = []
```
whoarethebritons: since we always use this for its length, would it be better to have a counter instead?

appscale/tools/agents/azure_agent.py
tmarballi: It shouldn't be but somehow there is a problem with the Azure SDK, which unless I specifically convert it to str, it always complains that 'Subscription ID should be only be a string'. I put that check in only when I started facing that issue. Its weird and I was not able to find a reason why.

appscale/tools/agents/ec2_agent.py
```diff
@@ -259,7 +266,8 @@ def get_params_from_args(self, args):
       self.PARAM_KEYNAME : args['keyname'],
       self.PARAM_STATIC_IP : args.get(self.PARAM_STATIC_IP),
       self.PARAM_ZONE : args.get('zone'),
-      'IS_VERBOSE' : args.get('verbose', False)
+      'IS_VERBOSE' : args.get('verbose', False),
+      'autoscale_agent' : False
```
whoarethebritons: Should this also be AUTOSCALE_AGENT for consistency even though its not used?

PR #605, opened 2017-03-23
catch AppControllerException for cluster terminate

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/2990/

### DIFF
--- a/appscale/tools/appscale_tools.py
+++ b/appscale/tools/appscale_tools.py
@@ -825,7 +825,7 @@ class AppScaleTools(object):
         AppScaleLogger.warn(e)
       else:
         AppScaleLogger.verbose(e, options.verbose)
-        if e is AppControllerException:
+        if type(e) is AppControllerException:
           response = raw_input(
             'AppScale may not have shut down properly, are you sure you want '
             'to continue terminating? (y/N) ')
--- a/appscale/tools/appscale_tools.py
+++ b/appscale/tools/appscale_tools.py
@@ -815,18 +815,28 @@ class AppScaleTools(object):
       RemoteHelper.terminate_virtualized_cluster(options.keyname,
                                                  options.clean,
                                                  options.verbose)
-    except (IOError, AppScaleException):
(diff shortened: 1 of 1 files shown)


appscale/tools/appscale_tools.py
```diff
@@ -818,6 +818,15 @@ def terminate_instances(cls, options):
     except (IOError, AppScaleException):
       # Don't fail if we cannot find the configuration.
       pass
+    except AppControllerException as e:
+      # If we are terminating a cloud infrastructure, we should log and
+      # continue.
+      if (infrastructure in InfrastructureAgentFactory.VALID_AGENTS and
+            options.terminate):
+        AppScaleLogger.warn(e)
```
jeanleonov: @whoarethebritons , is this error ignored for cloud mode because instances will be terminated (removed) later for cloud (so AppScale shouldn't necessarily be properly terminated there)?
Maybe then it makes sense to skip `RemoteHelper.terminate_virtualized_cluster` for cloud if terminate flag is set?

jeanleonov: No.. It isn't really sensible to not perform terminate of AppScale services for cloud, because at least `RemoteHelper.terminate_virtualized_cluster` also can do clean, if clean flag is specified.

whoarethebritons: @jeanleonov What happens here is if the tools can't talk to the AppController and they were going to terminate instances, we continue on to terminate those instances. We do still want to run terminate_virtualized_cluster if possible (because it would be a cleaner shutdown)

whoarethebritons:
@jeanleonov I think that AppScaleLogger only has warn and does not have an error. Also if a user does a down --clean or a down, that exception is not handled and will throw an error and make an error log. The only case this exception is meant to cover is a down --terminate where we cannot reach the head node's AC.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3013/

PR #607, opened 2017-03-29
change terminate to stopping during a down

whoarethebritons:
also added warning after a down (without terminate) is run on a cloud environment


whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3021

PR #611, opened 2017-04-06
Ensure an adequate setuptools version is installed

cdonati:
The msrestazure source distribution requires at least setuptools 8 to be installed.

### DIFF
--- a/debian/appscale_build.sh
+++ b/debian/appscale_build.sh
@@ -25,9 +25,10 @@ case ${DIST} in
         ;;
 esac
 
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3073/

cdonati:
@jeanleonov https://github.com/AppScale/appscale/blob/master/AppTaskQueue/appscale/__init__.py#L1 is one example. Last time I tried it, that was causing some serious breakage with setuptools 34 (pip wasn't even working). It may have been fixed with a newer setuptools.

PR #615, opened 2017-04-20
Azure Create Image

whoarethebritons:


appscale/tools/agents/azure_agent.py
```diff
@@ -399,19 +405,34 @@ def create_virtual_machine(self, credentials, network_client, network_id,
...
+                       create_option=DiskCreateOptionTypes.from_image,
+                       name=vm_network_name, vhd=virtual_hd)
+      storage_profile = StorageProfile(image_reference=image_ref,
+                                       os_disk=os_disk)
+    else:
+      image_hd = VirtualHardDisk(uri=parameters[self.PARAM_IMAGE_ID])
+      os_disk = OSDisk(os_type=os_type, caching=CachingTypes.read_write,
+                       create_option=DiskCreateOptionTypes.from_image,
+                       name=vm_network_name, vhd=virtual_hd, image=image_hd)
+      storage_profile = StorageProfile(os_disk=os_disk)
```
tmarballi: Could you pull out the os_disk and storage_profile from the if condition so that there is less duplication like this:

tmarballi: image_ref = None
    image_hd = None
    if re.search(".*:.*:.*:.*", azure_image_id):
      AppScaleLogger.log("Using publisher image {}".format(azure_image_id))
      image_ref_params = azure_image_id.split(":")
      image_ref = ImageReference(publisher=image_ref_params[0],
                                 offer=image_ref_params[1],
                                 sku=image_ref_params[2],
                                 version=image_ref_params[3])
    else:
      image_hd = VirtualHardDisk(uri=parameters[self.PARAM_IMAGE_ID])

    os_disk = OSDisk(os_type=os_type, caching=CachingTypes.read_write,
                     create_option=DiskCreateOptionTypes.from_image,
                     name=vm_network_name, vhd=virtual_hd, image=image_hd)
    storage_profile = StorageProfile(image_reference=image_ref,
                                     os_disk=os_disk)

appscale/tools/agents/azure_agent.py
```diff
@@ -1193,3 +1214,92 @@ def does_resource_group_exist(self, resource_group_name, resource_client):
     if resource_group_name in resource_group_names:
       return True
     return False
+
+  def create_image(self, instance_id, image_name, parameters, dest_container):
+    """ Creates an Azure virtual machine using the network interface created.
```
tmarballi: Update the doc string to create image.

cdonati: This docstring sounds like the functions creates a vm from an image.

appscale/tools/agents/azure_agent.py
```diff
@@ -1193,3 +1214,92 @@ def does_resource_group_exist(self, resource_group_name, resource_client):
...
+    vm_deallocate = compute_client.virtual_machines.deallocate(resource_group,
+                                                               instance_id)
+    # Wait until we have finished deallocating.
+    vm_deallocate.wait()
+
+    # Then we generalize it.
+    compute_client.virtual_machines.generalize(resource_group,
+                                               instance_id)
+    # Safety precaution sleep.
+    time.sleep(self.SLEEP_TIME)
```
cdonati: Is there a resource the function can can poll to determine if it can proceed?

appscale/tools/agents/azure_agent.py
```diff
@@ -1193,3 +1214,92 @@ def does_resource_group_exist(self, resource_group_name, resource_client):
...
+    resource_group = parameters[self.PARAM_RESOURCE_GROUP]
+    subscription_id = str(parameters[self.PARAM_SUBSCRIBER_ID])
+    compute_client = ComputeManagementClient(credentials, subscription_id)
+    storage_account = parameters[self.PARAM_STORAGE_ACCOUNT]
+
+    # First we must deallocate the VM we wish to capture.
+    vm_deallocate = compute_client.virtual_machines.deallocate(resource_group,
+                                                               instance_id)
+    # Wait until we have finished deallocating.
+    vm_deallocate.wait()
```
tmarballi: Maybe you know this already but the virtual machine operations like update/dellocate return a AzureOperationPoller instance which has methods like wait/done etc. The wait method here waits forever until a timeout is specified which we haven't here. It might be advisable to use a done instead (https://github.com/Azure/msrestazure-for-python/blob/4523c47a23a2cf2fc99a649d926e99aa47ce8d77/msrestazure/azure_operation.py#L490) which returns when the process is completed.

appscale/tools/agents/azure_agent.py
```diff
@@ -1193,3 +1214,92 @@ def does_resource_group_exist(self, resource_group_name, resource_client):
     if resource_group_name in resource_group_names:
       return True
     return False
+
+  def create_image(self, instance_id, image_name, parameters, dest_container):
```
cdonati: Does this function belong in the tools? It seems like appscale-testing would be a better place for it since the user does not need to perform this operation.

appscale/tools/agents/azure_agent.py
tmarballi: Same comment as before, here too you could assign the generalize operation to a result object and check the done() method on it instead of waiting an arbitrary sleep time.

appscale/tools/agents/azure_agent.py
whoarethebritons: @cdonati So I agree that the user shouldn't use this, but (as far as I can see) there isn't a way for the user to call this method and a large majority of it seems tied to the azure agent in my opinion.

whoarethebritons: My basis for having used this structure was the EC2 agent, which has a create_image function that appscale-testing uses.

whoarethebritons:
https://ocd.appscale.com:8080/view/All/job/Create-Azure-Image/39/

PR #616, opened 2017-04-26
Start AppController service directly

cdonati:
The controller will add itself to monit after starting.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3157/

PR #618, opened 2017-05-02
Extract AdminServer error messages properly

cdonati:
Previously, it would just report the error code instead of the message.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3214/

PR #621, opened 2017-05-17
Fix azure terminate to allow for downscaling specific instances

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/3243/

### DIFF
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -734,11 +734,7 @@ class AzureAgent(BaseAgent):
       for vmss in vmss_list:
         vm_list = compute_client.virtual_machine_scale_set_vms.list(
           resource_group, vmss.name)
-        vm_names = []
-        for vm in vm_list:
-          vm_names.append(vm.name)
-
-        if not vm_names:
+        if not any(True for _ in vm_list):
           thread = threading.Thread(
             target=self.delete_virtual_machine_scale_set, args=(
               compute_client, parameters, vmss.name))
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -742,29 +742,30 @@ class AzureAgent(BaseAgent):
     AppScaleLogger.log("Load balancer virtual machine(s) have been successfully "
                        "deleted")
 
-  def delete_virtual_machine_scale_set(self, compute_client, resource_group,
-                                       verbose, scale_set_name):
+  def delete_virtual_machine_scale_set(self, compute_client, parameters, vmss_name):
     """ Deletes the virtual machine scale set created from the specified
     resource group.
     Args:
         compute_client: An instance of the Compute Management client.
-        resource_group: The resource group name to use for this deployment.
-        verbose: A boolean indicating whether or not the verbose mode is on.
-        scale_set_name: The name of the virtual machine scale set to be deleted.
+        parameters: A dict, containing all the parameters necessary to
+          authenticate this user with Azure.
+        vmss_name: The name of the virtual machine scale set to be deleted.
     """
-    AppScaleLogger.verbose("Deleting Scale Set {} ...".format(scale_set_name), verbose)
+    resource_group = parameters[self.PARAM_RESOURCE_GROUP]
+    verbose = parameters[self.PARAM_VERBOSE]
+    AppScaleLogger.verbose("Deleting Scale Set {} ...".format(vmss_name), verbose)
     try:
       delete_response = compute_client.virtual_machine_scale_sets.delete(
-        resource_group,scale_set_name)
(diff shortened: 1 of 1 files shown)


appscale/tools/agents/azure_agent.py
```diff
@@ -700,81 +700,148 @@ def terminate_instances(self, parameters):
...
+      AppScaleLogger.log("Cleaning up any Scale Sets, if needed ...")
+      vmss_delete_threads = []
+      for vmss in vmss_list:
+        vm_list = compute_client.virtual_machine_scale_set_vms.list(
+          resource_group, vmss.name)
+        vm_names = []
+        for vm in vm_list:
+          vm_names.append(vm.name)
+
+        if not vm_names:
```
obino: Could we just check the size of vm_list instead of creating a new array?

appscale/tools/agents/azure_agent.py
```diff
@@ -700,81 +700,148 @@ def terminate_instances(self, parameters):
...
+      instance_id: The ID of the instance in the Scale Set to be deleted.
+    """
+    resource_group = parameters[self.PARAM_RESOURCE_GROUP]
+    verbose = parameters[self.PARAM_VERBOSE]
+    AppScaleLogger.verbose("Deleting Virtual Machine Instance {0} from Scale "
+      "Set {1} ...".format(instance_id, vmss_name), verbose)
+    result = compute_client.virtual_machine_scale_set_vms.delete(resource_group,
+                                                                 vmss_name,
+                                                                 instance_id)
+    resource_name = 'Virtual Machine Instance ' + instance_id
```
obino: Should Virtual Machine Instance be a constant? I assume it's a fixed construct from Azure?

tmarballi: Not really, so I have a common method for waiting until the resources are created/deleted. So I pass the resource name so that I can print in the logs which is the resource we are waiting on. I don't see any benefit in making it a constant because we would not really use it anywhere else. Unless you see it as neater?

appscale/tools/agents/azure_agent.py
tmarballi: The VM list we get from Azure is a Paged object and we don't have any methods we can use on it to get its size and it can only be iterated over to check the size by creating a new array.

appscale/tools/agents/azure_agent.py
```diff
@@ -700,81 +700,148 @@ def terminate_instances(self, parameters):
...
+      for delete_thread in vmss_vm_delete_threads:
+        delete_thread.join()
+
+      AppScaleLogger.log("Virtual machine(s) have been successfully downscaled.")
+      AppScaleLogger.log("Cleaning up any Scale Sets, if needed ...")
+      vmss_delete_threads = []
+      for vmss in vmss_list:
+        vm_list = compute_client.virtual_machine_scale_set_vms.list(
+          resource_group, vmss.name)
+        vm_names = []
```
cdonati: If you want to just check if there are no elements in an iterable without iterating through the whole thing, you can do:
```
if not any(True for _ in vm_list):
  [delete_scale_set]
```
`any` will stop on the first element, so it would prevent paging through a large scale set.

In cases where you really do need to exhaust the iterable, the `list()` function can be a nice shortcut.

appscale/tools/agents/azure_agent.py
```diff
@@ -700,81 +700,148 @@ def terminate_instances(self, parameters):
...
-    ss_instance_ids_to_delete = []
     for vmss in vmss_list:
       vm_list = compute_client.virtual_machine_scale_set_vms.list(
         resource_group, vmss.name)
       for vm in vm_list:
-        ss_instance_ids_to_delete.append(vm.name)
-      thread = threading.Thread(
-        target=self.delete_virtual_machine_scale_set, args=(
-          compute_client, resource_group, verbose, vmss.name))
+        delete_ss_instances.append(vm.name)
```
cdonati: If the entire scale set is being deleted, what is this list for?

appscale/tools/agents/azure_agent.py
tmarballi: Thats awesome, I did not know about that! Thanks, I will make that change :)

appscale/tools/agents/azure_agent.py
tmarballi: The list of instances to delete consists of both the scale set vms and the regular (load balancer) vms. I need this list of scale set vms so that I only give the remaining vms to delete to the delete_virtual_machine method. In this line over here: https://github.com/AppScale/appscale-tools/pull/621/files#diff-6b911255ca065691da775e67ef91b1e2R768

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/3259/
New Migration build: https://ocd.appscale.com:8080/job/Migration/1471/

PR #622, opened 2017-05-27
Handle lack of vendored dependencies in requests

cdonati:
This change was introduced in requests 2.16.


cdonati:
msrest depends on an older version of requests.

PR #626, opened 2017-06-14
Remove unused method

cdonati:
Now that undeploy uses the AdminServer, this method is no longer used.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3385/

PR #627, opened 2017-06-14
Ips layout redesign

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/3381/
https://ocd.appscale.com:8080/job/Daily%20Build/3384/ with AppScale/appscale-testing#242

### DIFF
--- a/appscale/tools/node_layout.py
+++ b/appscale/tools/node_layout.py
@@ -8,6 +8,7 @@ import yaml
 
 # AppScale-specific imports
 from agents.factory import InfrastructureAgentFactory
+from appscale_logger import AppScaleLogger
 from custom_exceptions import BadConfigurationException
 
 
@@ -102,16 +103,16 @@ class NodeLayout():
     """
     if not isinstance(options, dict):
       options = vars(options)
-    self.warning = None
+    self.master = None
     input_yaml = options.get('ips')
     if isinstance(input_yaml, str):
       with open(input_yaml, 'r') as file_handle:
         self.input_yaml = yaml.safe_load(file_handle.read())
     elif isinstance(input_yaml, dict):
       self.input_yaml = input_yaml
-      self.warning = "The AppScalefile is changing, the layout you are using " \
-                     "will be invalid soon. Please see our website for more " \
-                     "details."
+      AppScaleLogger.warn("The AppScalefile is changing, the layout you are "
+                          "using will be invalid soon. Please see our website "
+                          "for more details.")
     elif isinstance(input_yaml, list):
       self.input_yaml = input_yaml
     else:
@@ -265,7 +266,8 @@ class NodeLayout():
 
         if not node.is_valid():
           self.invalid(",".join(node.errors()))
-
+        if is_master:
+          self.master = node
         if self.infrastructure in InfrastructureAgentFactory.VALID_AGENTS:
           if not self.NODE_ID_REGEX.match(node.public_ip):
             self.invalid("{0} is not a valid node ID (must be node-id).".
@@ -322,10 +324,8 @@ class NodeLayout():
       for node in nodes:
         node.disk = self.disks.get(node.public_ip)
 
-    rep = self.is_database_replication_valid(nodes)
-
-    if not rep:
-      return rep
+    # Validate that database replication is valid.
+    self.validate_database_replication(nodes)
 
     self.nodes = nodes
     return True
@@ -351,26 +351,24 @@ class NodeLayout():
       'zookeeper': 0,
       'login': 0
     }
-    node_count = 1
     db_master_created = False
     tq_master_created = False
-    master_node = None
     login_found = False
     # Loop through the list of "node sets", which are grouped by role.
     for role, ips in self.input_yaml.iteritems():
       if len(ips) == 0:
-        self.invalid("Node amount cannot be zero.")
+        self.invalid("Node amount cannot be zero for role {}.".format(role))
       ips_list = ips if isinstance(ips, list) else [ips]
       test_ips = [self.is_cloud_ip(ip) for ip in ips_list]
-      cloud = all(test_ips) or not any(test_ips)
+      using_cloud_ids = any(test_ips)
 
       # Immediately fail if we have more than one node for master.
-      if role == 'master' and master_node is not None:
+      if role == 'master' and self.master:
         self.invalid("Only one master is allowed")
 
       # Create or retrieve the nodes from the node_hash.
-      nodes = [node_hash[ip] if ip in node_hash else AdvancedNode(ip, cloud)
-               for ip in ips_list]
+      nodes = [node_hash[ip] if ip in node_hash else \
+               AdvancedNode(ip, using_cloud_ids) for ip in ips_list]
 
       # Validate volume usage, there should be an equal number of volumes to
       # number of nodes.
@@ -383,10 +381,11 @@ class NodeLayout():
       # Add the defined roles to the nodes.
       for node in nodes:
         node.add_role(role)
-
+        if not node.is_valid():
+          self.invalid(",".join(node.errors()))
       # Check cases where a master is needed.
       if role == 'master':
-        master_node = nodes[0]
+        self.master = nodes[0]
       if role == 'database' and not db_master_created:
         nodes[0].add_db_role(True)
         db_master_created = True
@@ -407,13 +406,9 @@ class NodeLayout():
       # Update the node_hash with the modified nodes.
       node_hash.update({(node.public_ip, node) for node in nodes})
 
-      # Update node_count.
-      node_count += len(nodes)
-
-    # Dont need the hash any more, make a nodes list
-    nodes = node_hash.values()
-
-    self.distribute_unassigned_roles(nodes, role_count, master_node)
+    # Distribute unassigned roles and validate that certain roles are filled
+    # and return a list of nodes or raise BadConfigurationException.
+    nodes = self.distribute_unassigned_roles(node_hash.values(), role_count)
 
     if self.infrastructure in InfrastructureAgentFactory.VALID_AGENTS:
       if not self.min_vms:
@@ -421,9 +416,7 @@ class NodeLayout():
       if not self.max_vms:
         self.max_vms = len(nodes)
 
-    rep = self.is_database_replication_valid(nodes)
-    if not rep:
-      return rep
+    self.validate_database_replication(nodes)
 
     self.nodes = nodes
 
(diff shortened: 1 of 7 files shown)


whoarethebritons:
If you don't mind I'd like some opinions on f679af130724141e6aac8f8ca69c697f55d252aa as well

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+      if role == 'login':
+        login_found = True
+      elif role == 'login' and login_found:
+        self.invalid("Only one login is allowed")
 
-    if self.add_to_existing:
-      self.nodes = nodes
-      return self.valid()
+      # Update dictionary containing role counts.
+      role_count.update({(role, role_count.get(role, 0) + len(nodes))})
```
cdonati: Should this be `.update({role: role_count.get(role, 0) + len(nodes)})`? Also, since you're just updating one role, it might be clearer to do `role_count[role] += len(nodes)`

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+      # Update the node_hash with the modified nodes.
+      node_hash.update({(node.public_ip, node) for node in nodes})
```
cdonati: I'm not sure what's going on here. It looks like node_hash is a dictionary, but you're updating it with a set of tuples?

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
 
-    master_nodes = []
-    for node in nodes:
-      if node.is_role('shadow'):
-        master_nodes.append(node)
+      # Update node_count.
+      node_count += len(nodes)
```
cdonati: What's this used for?

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
-    # need exactly one master
-    if len(master_nodes) == 0:
-      return self.invalid("No master was specified")
-    elif len(master_nodes) > 1:
-      return self.invalid("Only one master is allowed")
+    # Dont need the hash any more, make a nodes list
+    nodes = node_hash.values()
 
-    master_node = master_nodes[0]
+    self.distribute_unassigned_roles(nodes, role_count, master_node)
```
cdonati: It's easier to maintain functions that don't have side effects. Returning a new nodes structure would make it easier to tell what's being changed, and it would keep state separated.

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
 
-    login_nodes = []
-    for node in nodes:
-      if node.is_role('login'):
-        login_nodes.append(node)
+    if self.infrastructure in InfrastructureAgentFactory.VALID_AGENTS:
+      if not self.min_vms:
+        self.min_vms = len(nodes)
+      if not self.max_vms:
+        self.max_vms = len(nodes)
```
cdonati: Do we rely on these being set to prevent scaling? It might be cleaner to not pass infrastructure instead.

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+    node_count = 1
+    db_master_created = False
+    tq_master_created = False
+    master_node = None
+    login_found = False
+    # Loop through the list of "node sets", which are grouped by role.
+    for node_set in self.input_yaml:
+      # If the key nodes is mapped to an integer it should be a cloud
+      # deployment.
+      cloud = True if isinstance(node_set.get('nodes'), int) else False
```
cdonati: `cloud = isinstance(node_set.get('nodes'), int)`. I would reconsider the name since It can be a cloud deployment without nodes being an integer.

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+    node_hash = {}
+    role_count = {
+      'appengine': 0,
+      'master': 0,
+      'memcache': 0,
+      'taskqueue': 0,
+      'zookeeper': 0,
+      'login': 0
+    }
+    node_count = 1
```
cdonati: It would be easier to reason about this if it started at 0. You can add 1 when formatting it.

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+      if len(ips_list) == 0:
+        self.invalid("Node amount cannot be zero.")
+
+      # Get the roles.
+      role_or_roles = node_set.get('roles')
+      roles = role_or_roles if isinstance(role_or_roles, list) else \
+        [role_or_roles]
+
+      # Immediately fail if we have more than one node for master.
+      if 'master' in roles and master_node and len(ips_list) > 1:
```
cdonati: What's the reason for the `and len(ips_list) > 1` check?

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+                         for role in roles})
+      # Update the node_hash with the modified nodes.
+      node_hash.update({(node.public_ip, node) for node in nodes})
 
-    memcache_count = 0
-    for node in nodes:
-      if node.is_role('memcache'):
-        memcache_count += 1
+      # Update node_count.
+      node_count += len(nodes)
```
cdonati: If a node existed in a previous set, should it contribute to the count?

appscale/tools/node_layout.py
```diff
@@ -590,10 +597,46 @@ def is_database_replication_valid(self, nodes):
...
+  def distribute_unassigned_roles(self, nodes, role_count, master_node):
+    """ Distributes roles that were not defined by user.
+    
+    Args:
+      nodes: The list of nodes.
+      role_count: A dict containing roles mapped to their count.
+      master_node: The master node of the deployment.
+    """
+    # Check if a master node was specified.
+    if role_count.pop('master', 0) == 0:
```
cdonati: Why does role_count need to be altered?

appscale/tools/node_layout.py
```diff
@@ -602,15 +645,11 @@ def generate_cloud_layout(self):
...
-    for i in xrange(num_slaves):
-      servers.append("node-{0}".format(i+2))
-
-    layout['servers'] = servers
-    return layout
-
+    return {
+      'controller': "node-1",
+      'servers': ["node-{}".format(i)\
+                  for i in xrange(2, self.min_vms - 1)]
```
cdonati: This would result in one fewer machines than `min_vms` if `min_vms` > 1.

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+      # Validate volume usage, there should be an equal number of volumes to
+      # number of nodes.
+      if self.disks:
+        self.validate_disks(nodes)
 
-    for node in nodes:
-      if not node.is_valid():
-        return self.invalid(",".join(node.errors()))
+        for node in nodes:
+          node.disk = self.disks.get(node.public_ip)
```
whoarethebritons: the public_ip is always assigned currently and can also be node-1

appscale/tools/node_layout.py
whoarethebritons: this is used to determine node-id for nodes, it could be moved though.

appscale/tools/node_layout.py
whoarethebritons: can't have more than one master

appscale/tools/node_layout.py
whoarethebritons: not possible in the case where node_count is actually used, node count is used for the `nodes: 3` case to expand it into node-1, node-2, node-3

appscale/tools/node_layout.py
whoarethebritons: master doesn't need to be a part of the loop for distributing roles

appscale/tools/node_layout.py
whoarethebritons: yes but the controller already takes one of those machines.

appscale/tools/node_layout.py
```diff
@@ -367,202 +317,259 @@ def is_valid_simple_format(self):
...
+    # Loop through the list of "node sets", which are grouped by role.
     for role, ips in self.input_yaml.iteritems():
-      
-      if isinstance(ips, str):
-        ips = [ips]
+      if len(ips) == 0:
+        self.invalid("Node amount cannot be zero.")
+      ips_list = ips if isinstance(ips, list) else [ips]
+      test_ips = [self.is_cloud_ip(ip) for ip in ips_list]
+      cloud = all(test_ips) or not any(test_ips)
```
whoarethebritons: test_ips is a list of boolean values because is_cloud_ip returns a boolean value, so they should all be true if we were given node-id and none should be true if we were given ip addresses. This does mean we can't do what we're currently able to do where we can use node-id and ip addresses at the same time. If we want to keep supporting the ability to use them both together it could just be `cloud = any(test_ips)`

appscale/tools/node_layout.py
```diff
@@ -96,16 +97,23 @@ def __init__(self, options):
...
       options = vars(options)
-
+    self.warning = None
     input_yaml = options.get('ips')
     if isinstance(input_yaml, str):
       with open(input_yaml, 'r') as file_handle:
         self.input_yaml = yaml.safe_load(file_handle.read())
     elif isinstance(input_yaml, dict):
       self.input_yaml = input_yaml
+      self.warning = "The AppScalefile is changing, the layout you are using " \
```
whoarethebritons: Yes I wasn't sure if the AppScaleLogger should be imported into NodeLayout, but I guess RemoteHelper does it.

appscale/tools/node_layout.py
whoarethebritons: agh, yes it should, I got my parantheses mixed up. the reason I was doing update was so that I didn't have to initialize the dictionary with roles besides the necessary ones.

appscale/tools/node_layout.py
whoarethebritons: actually upon looking at it, its left over from a copy paste its not used in node format.

appscale/tools/node_layout.py
whoarethebritons: its mostly from the node-1 convention but since the user wouldn't ever see it it could just start at 0 and not have the one added.

appscale/tools/node_layout.py
whoarethebritons: I'm not sure, its open to debate. That's what was currently there.

appscale/tools/node_layout.py
cdonati: The first part of the condition should be enough for that, right? The extra condition just prevents the exception from being thrown if the size of the current ips_list is greater than 1. For example, if master was already defined in a different node_set and it is defined again in this set and the size of the current ips_list is greater than 1, the master_node will just get overwritten.

appscale/tools/node_layout.py
whoarethebritons: This case is going to be deprecated soon. Additionally this is the current `node-1: disk-1` where the public ip actually is the key.

appscale/tools/node_layout.py
cdonati: What if someone mixes nodes as integers, specifies nodes, and re-uses nodes? Maybe that's not allowed since we would probably be re-using nodes in that case.

appscale/tools/node_layout.py
cdonati: If that's the case, you can ignore it in the loop. Altering a data structure that's been given can be surprising to the calling function.

appscale/tools/node_layout.py
cdonati: If `min_vms` is 2, the controller would get one, but there would be nothing in `servers`.

appscale/tools/node_layout.py
cdonati: Cool. It makes sense to keep it, then.

appscale/tools/node_layout.py
whoarethebritons: I think I should not allow being able to specify node-id since I'm using them for the integers. If you think about it there's really no need for people to specify node-1 since they'd be using it to put multiple roles and now its role based.

cdonati: That makes sense. It sounds like it would simplify things a bit, too.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3510/

PR #630, opened 2017-06-22
Added support for m1.large EC2 machine type

jeanleonov:
Before getting any approvals i'd like to check with you if we need to add other instance types supported by ec2? What was a purpose for skipping some instance type in initial implementation?


obino:
@jeanleonov we cannot use m1.large on EC2 itself. This the error we get A client error (InvalidParameterCombination) occurred when calling the RunInstances operation: Non-Windows instances with a virtualization type of 'hvm' are currently not supported for this instance type.

Unless we can differentiate between euca and AWS, we should not have old instance type in the listing it seems.

PR #631, opened 2017-07-03
Fix gathering logs

jeanleonov:
- Use absolute pass when copying logs.
 - Handle verbose flag for logs command.

### DIFF
--- a/appscale/tools/appscale_tools.py
+++ b/appscale/tools/appscale_tools.py
@@ -486,7 +486,7 @@ class AppScaleTools(object):
       os.mkdir(local_dir)
 
       if ip == login_host:
-        os.symlink(local_dir, os.path.join(location, "login-host"))
+        os.symlink(local_dir, os.path.join(location, "load-balancer"))
 
       for log_path in log_paths:
         sub_dir = local_dir
--- a/appscale/tools/appscale_tools.py
+++ b/appscale/tools/appscale_tools.py
@@ -448,6 +448,7 @@ class AppScaleTools(object):
     if os.path.exists(options.location):
       raise AppScaleException("Can't gather logs, as the location you " + \
         "specified, {0}, already exists.".format(options.location))
+    location = os.path.abspath(options.location)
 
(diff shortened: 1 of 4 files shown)


appscale/tools/appscale_tools.py
```diff
@@ -480,9 +482,12 @@ def gather_logs(cls, options):
     failures = False
     for ip in all_ips:
       # Get the logs from each node, and store them in our local directory
-      local_dir = "{0}/{1}".format(options.location, ip)
+      local_dir = os.path.join(location, ip)
       os.mkdir(local_dir)
 
+      if ip == login_host:
+        os.symlink(local_dir, os.path.join(location, "login-host"))
```
obino: Since login-host is not a role, and we don't use anywhere else, could we change this to something a user can relate more? Is this the load-balancer perhaps?

appscale/tools/appscale_tools.py
```diff
@@ -480,9 +482,12 @@ def gather_logs(cls, options):
     failures = False
     for ip in all_ips:
       # Get the logs from each node, and store them in our local directory
-      local_dir = "{0}/{1}".format(options.location, ip)
+      local_dir = os.path.join(location, ip)
       os.mkdir(local_dir)
 
+      if ip == login_host:
+        os.symlink(local_dir, os.path.join(location, "load-balancer"))
```
cdonati: I don't think this will work since that location (local_dir) already exists.

jeanleonov: [os.symlink](https://docs.python.org/2/library/os.html#os.symlink) takes two arguments: `source` and `link_name`. Source should exist before creating link.

cdonati: You're right. I'm not sure what I was thinking.

jeanleonov:
https://ocd.appscale.com:8080/job/Tools-test/2195/

jeanleonov:
https://ocd.appscale.com:8080/job/Tools-test/2196/

PR #632, opened 2017-07-05
Appscale stats command for tools

Maksym-Us:
add stats command to print node, process and proxy statistics.


appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,626 @@
...
+  last_5 = loadavg["last_5min"]
+  last_15 = loadavg["last_15min"]
+  return "{} / {} / {}".format(
+    last_1 if last_1 < limit_value else get_marked(last_1, "red"),
+    last_5 if last_5 < limit_value else get_marked(last_5, "red"),
+    last_15 if last_15 < limit_value else get_marked(last_15, "red")
+  )
+
+
+def render_partitions(partitions, verbose=False):
```
jeanleonov: It will always get verbose flag, no need to provide default value here.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,655 @@
...
+        n["roles"] for n in cluster_stats if n["private_ip"] == ip
+      )
+    for ip in all_private_ips
+  }
+
+  return roles_data
+
+
+def get_node_stats(raw_node_stats, all_roles,
+                   specified_roles=None, verbose=False):
```
jeanleonov: This function is called only from show_stats function and there all arguments are specified. So there is no need to assign default values to `specified_roles` and `verbose` arguments.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,655 @@
...
+      unique_memory / 1024 / 1024,  # row[2] from B to MB
+      cpu_percent,
+      avg_cpu_per_process,
+      cpu_percent_cpu_count
+    ])
+
+  return process_stats_headers, process_stats
+
+
+def get_proxy_stats(raw_proxy_stats, verbose=False, apps_filter=False):
```
jeanleonov: Default values are redundant.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,655 @@
...
+
+  partitions_info = [
+    "'{}': {}%".format(part[0], part[1])
+    if part[1] < 90
+    else get_marked("'{}': {}%".format(part[0], part[1]), "red")
+    for part in part_list
+  ]
+
+  if not verbose:
+    if len(partitions_info) > 3:
```
jeanleonov: Alternatively you could write `if not verbose and len(partitions_info) > 3:`. JFYI (Just for your information)

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,655 @@
...
+  proxy_stats = []
+  proxy_groups = []
+  unique_proxies = {}
+
+  for node in raw_proxy_stats.itervalues():
+    proxy_groups += node["proxies_stats"]
+
+  for node in proxy_groups:
+    if apps_filter:
+      if "application" != node["unified_service_name"]:
```
jeanleonov: alternative:
`if apps_filter and "application" != node["unified_service_name"]:`

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,655 @@
...
+        continue
+
+    service_name_id = node["unified_service_name"]\
+                      + (" | " + node["application_id"]
+                         if node["application_id"]
+                         else "")
+
+    if not service_name_id in unique_proxies:
+      unique_proxies[service_name_id] = {}
+      unique_proxies[service_name_id]["servers_count"] = set()
```
jeanleonov: Getting `unique_proxies[service_name_id]` so many times can be expensive. I don't think it'll make show stats command slow.. but it still can be improved like this:
```
proxy_summary = unique_proxies.get(service_name_id)
if not proxy_summary:
  unique_proxies[service_name_id] = proxy_summary = {}
  proxy_summary["servers_count"] = set()
  ...
```

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,655 @@
...
+  ]
+
+  proxy_stats_headers_verbose = [
+    "SERVICE (ID)", "SERVERS", "REQ RATE / REQ TOTAL", "5xx / 4xx",
+    "BYTES IN / BYTES OUT", "SESSION CUR / QUEUE CUR", "QTIME / RTIME"
+  ]
+
+  proxy_stats = []
+  proxy_groups = []
+  unique_proxies = {}
```
jeanleonov: Hmm.. I'm afraid I asked you to calculate summary for proxies.. I'm not sure)
I guess that proxies stats are already unique. There is no need to group them. Let's discuss it in office.

Maksym-Us: I wrote it in case if there are a few load balancer nodes.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,626 @@
...
+      summary = True
+      top = None
+      if options.verbose:
+        summary = False
+      if "name" in options.order_processes:
+        sort = 0 if summary else 1
+      elif "mem" in options.order_processes:
+        sort = 2 if summary else 2
+      elif "cpu" in options.order_processes:
+        sort = 3 if summary else 3
```
Maksym-Us: In case when you set "choices" property in argparse library it returns a list but not a string.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+  Args:
+    process_stats: A list of process statistics.
+    column: An int representing a column number the list should be sorted by.
+    top: An int representing a process count to be printed.
+    reverse: A boolean to reverse or not reverse sorted list.
+
+  Returns:
+    A list of top processes sorted by specified column.
+  """
+  if top <= 0:
```
cdonati: I would avoid using possible value as an exceptional case. I don't think a user would ever specify `--top 0`, but I also don't think it would be intuitive for the tools to print everything if that value _was_ explicitly specified.

In this case, it would be conventional to just use `None` to signify that the user did not specify an argument by just removing the `default` parameter on this argument.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+
+  Returns:
+    A list sorted by specified column.
+  """
+  return sorted(proxy_stats, key=lambda p: p[column], reverse=False)
+
+
+def get_roles(keyname):
+  """
+  Obtaines roles for each ip from AppControllerClient.
```
cdonati: Obtains

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+
+  Returns:
+    A dict in which each key is an ip and value is a role list.
+  """
+  login_host = LocalState.get_login_host(keyname=keyname)
+  login_acc = AppControllerClient(
+    host=login_host,
+    secret=LocalState.get_secret_key(keyname)
+  )
+  all_private_ips = login_acc.get_all_private_ips()
```
cdonati: I'm not sure why you need a list of IPs if that info is included in the stats.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+  cluster_stats = login_acc.get_cluster_stats()
+  roles_data = {
```
cdonati: For example, this could just be `{node['ip']: node['role'] for node in cluster_stats}`.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+  return roles_data
+
+
+def get_node_stats(raw_node_stats, all_roles, specified_roles, verbose):
+  """
+  Obtaines useful information from node statistics and returns:
+  PRIVATE IP, AVAILABLE MEMORY, LOADAVG, PARTITIONS USAGE, ROLES values.
+
+  Args:
+    raw_node_stats: A dist in which each key is an ip and value is a dict
```
cdonati: dict

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+      )
+    for ip in all_private_ips
+  }
+
+  return roles_data
+
+
+def get_node_stats(raw_node_stats, all_roles, specified_roles, verbose):
+  """
+  Obtaines useful information from node statistics and returns:
```
cdonati: Obtains

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+  return node_stats_headers, node_stats
+
+
+def get_process_stats(raw_process_stats):
+  """
+  Obtains useful information from process statistics and returns:
+  PRIVATE IP, MONIT NAME, UNIQUE MEMORY (MB), CPU (%) values.
+
+  Args:
+    raw_process_stats: A dist in which each key is an ip and value is a dict
```
cdonati: dict

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+    A list of process statistics headers.
+    A list of process statistics.
+  """
+  process_stats_headers = [
+    "PRIVATE IP", "MONIT NAME", "UNIQUE MEMORY (MB)", "CPU (%)"
+  ]
+
+  process_stats = []
+  for ip, node in raw_process_stats.iteritems():
+    stat = node.get("processes_stats")
```
cdonati: `get` usually indicates that the function should proceed if the key doesn't exist. You could either do `node.get('processes_stats', [])` (since an iterable is expected below) or `node['processes_stats']` if an exception should be thrown.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+
+
+def get_summary_process_stats(raw_process_stats, raw_node_stats):
+  """
+  Obtains useful information from summary process statistics and returns:
+  SERVICE (ID), INSTANCES, UNIQUE MEMORY SUM (MB),
+  CPU SUM (%), CPU PER 1 PROCESS (%), CPU PER 1 CORE (%) values.
+
+  Args:
+    raw_process_stats: A dist in which each key is an ip and value is a dict
```
cdonati: dict

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+    for p in process_group:
+      unique_memory += int(p["memory"]["unique"])
+      cpu_percent += float(p["cpu"]["percent"])
+      if p["children_num"] > 0:
+        child = p["children_stats_sum"]
+        unique_memory += int(child["memory"]["unique"])
+        cpu_percent += float(child["cpu"]["percent"])
+
+    avg_cpu_per_process = cpu_percent / len(process_group)
+    cpu_percent_cpu_count = cpu_percent / cpu_count
```
cdonati: I'm not sure this is the best way to determine CPU usage / core since our processes don't account for all the running processes on the machine. For example, if a user is running their own CPU-intensive process alongside AppScale, this value could be a lot lower than the actual CPU usage / core on that machine.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+
+
+def get_proxy_stats(raw_proxy_stats, verbose, apps_filter):
+  """
+  Obtains useful information from proxy statistics and returns:
+  SERVICE (ID), UNIQUE MEMORY SUM (MB), CPU PER 1 PROCESS (%),
+  CPU PER 1 CORE (%), CPU SUM (%) values.
+
+  Args:
+    raw_proxy_stats: A dist in which each key is an ip and value is a dict
```
cdonati: dict

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+      of useful information.
+    verbose: A boolean - verbose or not verbose mode.
+    apps_filter: A boolean - show all services or applications only.
+
+  Returns:
+    A list of proxy statistics headers.
+    A list of proxy statistics.
+  """
+  proxy_stats_headers = [
+    "SERVICE | ID", "SERVERS", "REQ RATE / REQ TOTAL", "5xx / 4xx", "QUEUE CUR"
```
cdonati: "REQ RATE / REQ TOTAL" is an unusual way of displaying these values because it implies that it is a fraction.

appscale/tools/appscale_stats.py
```diff
@@ -0,0 +1,657 @@
...
+  """
+  Prints a list of statistics with specified headers.
+
+  Args:
+    table_name: A string representing a name of table.
+    headers: A list of statistics headers.
+    data: A list of statistics.
+  """
+  table = tabulate(tabular_data=data, headers=headers, tablefmt='simple',
+               floatfmt=".1f", numalign="right", stralign="left")
```
cdonati: align with start of args above

appscale/tools/parse_args.py
```diff
@@ -375,6 +375,32 @@ def add_allowed_flags(self, function):
...
+        help="print nodes with specified role(s)")
+      self.parser.add_argument('--order-processes', '-o',
+        nargs=1,
+        choices=['cpu', 'mem', 'name'],
+        default=['cpu'],
+        help="the arguments provide processes statistics order by")
+      self.parser.add_argument('--top',
+        type=int,
+        default=0,
+        help="a number of top-processes to be printed")
```
cdonati: I would recommend `--limit` here. "the number of processes to print"

appscale/tools/parse_args.py
```diff
@@ -375,6 +375,32 @@ def add_allowed_flags(self, function):
...
+        nargs='+',
+        choices=['nodes', 'processes', 'proxies'],
+        default=['nodes', 'proxies'],
+        help="the list of entities that statistics should be printed")
+      self.parser.add_argument('--roles', '-r',
+        nargs='*',
+        default=[],
+        help="print nodes with specified role(s)")
+      self.parser.add_argument('--order-processes', '-o',
+        nargs=1,
```
Maksym-Us: Yes, this only allows one argument, but if I specify a choice list to the argument and string as a default type, it will return string in case when I  do not write this argument in a command line and return list in case when I do write this argument.
For example:
"... --show processes ..." returns 'cpu'
"... --show processes -o cpu ..." returns ['cpu']
So, I need to specify list as a default type.

appscale/tools/parse_args.py
```diff
@@ -375,6 +375,32 @@ def add_allowed_flags(self, function):
...
         help="the keypair name to use")
+    elif function == "appscale-show-stats":
+      self.parser.add_argument('--keyname', '-k',
+        default=self.DEFAULT_KEYNAME,
+        help="the keypair name to use")
+      self.parser.add_argument('--show',
+        nargs='+',
+        choices=['nodes', 'processes', 'proxies'],
+        default=['nodes', 'proxies'],
+        help="the list of entities that statistics should be printed")
```
Maksym-Us: "I would also suggest --type instead of --show, but I don't feel strongly about it."
Is it appropriate in case when an user can specify a few types there?

cdonati: Good point. `--types` is probably better.

appscale/tools/parse_args.py
cdonati: My point was that you could remove `nargs` altogether to store the argument as a string. You can set `default='cpu'`.

Maksym-Us: Oh, thank you!

appscale/tools/appscale_stats.py
Maksym-Us: Okay, but if I clarify that I show only AppScale processes?
I mean, the table name will be "APPSCALE PROCESS STATISTICS" instead of "PROCESS STATISTICS".

appscale/tools/parse_args.py
Maksym-Us: I think `--top` is more suitable, because it chooses the most consumed processes.

appscale/tools/appscale_stats.py
cdonati: Sure, I think that's fine.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4717/

PR #638, opened 2017-07-17
Fix path length calculation

cdonati:
`os.path.split` always returns just two elements, so it's not useful for determining path length.

### DIFF
--- a/appscale/tools/utils.py
+++ b/appscale/tools/utils.py
@@ -54,24 +54,6 @@ def config_from_zip(file_name, zip_location):
       return config_file.read()
 
 
-def path_length(path):
-  """ Gets the length of a file system path.
-
-  Args:
-    path: A string specifying the location of a file.
-  Returns:
-    An integer specifying the length of the path.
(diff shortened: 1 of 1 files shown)


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/3576/

appscale/tools/utils.py
```diff
@@ -91,7 +73,7 @@ def config_from_dir(file_name, source_path):
 
   shortest_path = candidates[0]
   for candidate in candidates:
-    if path_length(candidate) < path_length(shortest_path):
+    if len(candidate.split(os.sep)) < len(shortest_path.split(os.sep)):
       shortest_path = candidate
```
jeanleonov: As an alternative:
`shortest_path = min(len(path.split(os.sep)) for path in candidates)`

cdonati: Wouldn't that just give me the length of the shortest path and not the shortest path itself?

PR #641, opened 2017-07-28
Fix unit tests

cdonati:
The symlink call should be mocked.


cdonati:
https://ocd.appscale.com:8080/job/Create-Eucalyptus-Image/4639/

PR #642, opened 2017-08-01
Test updated gather logs

jeanleonov:
Fixed a bug with test which had to be updated after changing gathering tests.
https://ocd.appscale.com:8080/job/Daily%20Build/3650/


test/test_appscale_gather_logs.py
```diff
@@ -107,6 +107,10 @@ def test_appscale_in_two_node_virt_deployment(self):
     os.should_receive('mkdir').with_args('/tmp/foobaz/public2/cassandra')
     os.should_receive('mkdir').with_args('/tmp/foobaz/public2/rabbitmq')
 
+    # fake creation of symlink to shadow's logs
+    os.should_receive('symlink').with_args('/tmp/foobaz/public1',
```
cdonati: It's strange that the line from https://github.com/AppScale/appscale-tools/pull/641 doesn't show in the diff here. Anyway, your fix is better.

PR #643, opened 2017-08-01
Add instance ID only when a private IP is assigned to the instance.

tmarballi:
Previously, the instance IDs were added without checking if the instances had been assigned private IPs, which in one case crashed the controller because one of the instances came back without an IP associated with it. This PR fixes that and makes sure the instance has a private IP before adding it to the list.

### DIFF
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -287,14 +287,15 @@ class AzureAgent(BaseAgent):
           list_virtual_machine_scale_set_vm_network_interfaces(resource_group,
                                                                vmss.name,
                                                                vm.instance_id)
-        ip_config_private_ips = []
+        ip_config_private_ip = None
         for network_interface in network_interface_list:
           for ip_config in network_interface.ip_configurations:
-            ip_config_private_ips.append(ip_config.private_ip_address)
+            ip_config_private_ip = ip_config.private_ip_address
+            break
 
-        if ip_config_private_ips:
-          public_ips.append(ip_config_private_ips[0])
-          private_ips.append(ip_config_private_ips[0])
+        if ip_config_private_ip:
+          public_ips.append(ip_config_private_ip)
+          private_ips.append(ip_config_private_ip)
           instance_ids.append(vm.name)
 
     return public_ips, private_ips, instance_ids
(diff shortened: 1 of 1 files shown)


appscale/tools/agents/azure_agent.py
```diff
@@ -283,14 +283,16 @@ def describe_instances(self, parameters, pending=False):
...
-      for network_interface in network_interface_list:
-        for ip_config in network_interface.ip_configurations:
-          public_ips.append(ip_config.private_ip_address)
-          private_ips.append(ip_config.private_ip_address)
+        network_interface_list = network_client.network_interfaces. \
+          list_virtual_machine_scale_set_vm_network_interfaces(resource_group,
+                                                               vmss.name,
+                                                               vm.instance_id)
+        for network_interface in network_interface_list:
+          for ip_config in network_interface.ip_configurations:
```
cdonati: These for loops make it seem possible for the VM to be added to the list multiple times. I think it would be clearer to ensure that the appending only happens at most once for each VM.

For example, you could generate a list of IP addresses associated with a VM and choose the first one (or continue to the next VM if none exist).

This would prevent issues in case Azure ever decides to start automatically giving machines multiple network interfaces or something.

appscale/tools/agents/azure_agent.py
```diff
@@ -283,14 +283,19 @@ def describe_instances(self, parameters, pending=False):
...
-          public_ips.append(ip_config.private_ip_address)
-          private_ips.append(ip_config.private_ip_address)
+        network_interface_list = network_client.network_interfaces. \
+          list_virtual_machine_scale_set_vm_network_interfaces(resource_group,
+                                                               vmss.name,
+                                                               vm.instance_id)
+        ip_config_private_ips = []
+        for network_interface in network_interface_list:
+          for ip_config in network_interface.ip_configurations:
+            ip_config_private_ips.append(ip_config.private_ip_address)
```
whoarethebritons: if we only take the first one, we could change this to be:
```
ip_config_private_ip = None
for network_interface in network_interface_list:
  for ip_config in network_interface.ip_configurations:
    ip_config_private_ip = ip_config.private_ip_address
    break
  if ip_config_private_ip:
    break
```

right?

tmarballi: Yes that makes total sense, I will change it.

PR #653, opened 2017-09-29
Give GCE disk a more unique name

cdonati:
If the group name is long enough, there may not be enough characters for a unique time stamp.


cdonati:
Demo:
master fails with "resource already exists": https://ocd.appscale.com:8080/job/Hawkeye-Test-GCE-Image/186/
branch proceeds to tests: https://ocd.appscale.com:8080/job/Hawkeye-Test-GCE-Image/188/

PR #657, opened 2017-10-09
Create friendly navigation symlinks in gathered logs

jeanleonov:


appscale/tools/utils.py
```diff
@@ -120,3 +120,12 @@ def queues_from_xml(contents):
...
   return queues
+
+
+def mkdir(dir_path):
+  """ Creates a directory.
+
+  Args:
+    dir_path: The path to create.
+  """
+  return os.system("mkdir -p " + dir_path)
```
cdonati: Does this need to be a system call? It would make the package more difficult to port in the future.

appscale/tools/appscale_tools.py
```diff
@@ -482,35 +491,38 @@ def gather_logs(cls, options):
...
+
+      # Create symlinks for easier navigation in gathered logs
+      node_info = nodes_dict.get(public_ip)
+      if node_info:
+        private_ip = "{}-private-IP".format(node_info["private_ip"])
+        os.symlink(local_dir, os.path.join(location, private_ip))
+        for role in node_info['jobs']:
+          role_dir = os.path.join(location, "{}-nodes".format(role))
+          utils.mkdir(role_dir)
+          os.symlink(local_dir, os.path.join(role_dir, private_ip))
```
cdonati: These should probably be relative rather than absolute links. Otherwise, if a user creates a tarball and unpacks it elsewhere, the symlinks will be pointing to an absolute location from the original file system.

appscale/tools/utils.py
obino: Wouldn't it be better to use os.mkdir or os.makedirs? We should also check if we get a permissions denied, or already exists as a simple checks.

jeanleonov:
@cdonati , your comment about jenkins artifacts points to important issue. Current size of all build artifacts is ~380 GB. If we multiply it in even 2 times (in reality navigation redundancy will multiply logs size in 4-8 times) it will be huge number.

In order to avoid problems with disk space on jenkins I'm going to make creation of symlinks optional. Role-PrivateIP-PublicIP mapping I'll write to yaml file and save it to logs directory.

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4202/

jeanleonov:
https://ocd.appscale.com:8080/job/Daily%20Build/4208/

PR #660, opened 2017-10-11
Do not use regex to parse appengine-web.xml

cdonati:
Previously, a commented out application tag could be chosen as the project ID.


whoarethebritons:
So will this prevent us from grabbing any commented out XML?

obino:
@whoarethebritons the unit tests does exactly that check :)

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4097/

PR #667, opened 2017-11-03
Fix broken Context Manager reference

cdonati:
The extractfile method does not support the typical context manager syntax for working with the file object.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4435/

PR #669, opened 2017-11-09
Attach disk for EC2

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/4295/
https://ocd.appscale.com:8080/job/Migration/2297/


whoarethebritons:
@scragraham it does need a demo, but it also needs to be re-reviewed because of the last commit so I'll send out a message.

PR #671, opened 2017-11-10
Add support for Jessie

cdonati:


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4315/

appscale/tools/remote_helper.py
```diff
@@ -436,9 +430,11 @@ def enable_root_login(cls, host, keyname, infrastructure, is_verbose):
         raise exception
 
     # Amazon EC2 rejects a root login request and tells the user to log in as
-    # the ubuntu user, so do that to enable root login.
-    if re.search(cls.LOGIN_AS_UBUNTU_USER, output):
-      cls.merge_authorized_keys(host, keyname, 'ubuntu', is_verbose)
+    # a different user, so do that to enable root login.
+    error_prefix = 'Please login as the user'
+    if error_prefix in output:
+      user = output.split(error_prefix)[1].split()[0][1:-1]
```
jeanleonov: Wouldn't it better to use regex pattern like
```
'Please login as the user \"(?P<user>\w+)\" rather than the user \"\w+\".'
```
Having explicit patter would help to check if we're asked to login as another user, find what user should we use and in the same time it would make clearer code.

cdonati: Yes, I think that's clearer.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/4393/

PR #674, opened 2018-02-08
Error handling for azure agent autoscaling

whoarethebritons:
This branch is a result of issues we have noticed with autoscaling in Azure on master. It attempts to cover the following cases:

- Failure to create instances due to reaching quota or other issue
- Failure to delete instances and confirmation of that failure
- Failures caused by paging through Azure Resources where each is a request and we hit the request limit.

`run_and_return_exception`, `raise_if_error`, and `convert_to_list` are meant to address these issues so that we can capture the `CloudError` that is thrown and return it if we're running within a thread or raise it if we are not running within a thread. Any errors captured in threads are raised after the joins so that the caller can handle the `AgentRuntimeException`

This branch requires a change on the appscale side as well so that the InfrastructureManager can handle the `AgentRuntimeException` and also base_agent is modified here so it could be removed to remove confusion about that exception. I'm working on unit tests for the appscale side so that will be up shortly.

I'd like to get reviews before posting the demo in case I have to rerun it but I have a demo ready.

### DIFF
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -431,13 +431,17 @@ class AzureAgent(BaseAgent):
                      name=vm_network_name, vhd=virtual_hd, image=image_hd)
     storage_profile = StorageProfile(image_reference=image_ref,
                                      os_disk=os_disk)
-    compute_client.virtual_machines.create_or_update(
-      resource_group, vm_network_name, VirtualMachine(location=zone,
-                                                      os_profile=os_profile,
-                                                      hardware_profile=hardware_profile,
-                                                      network_profile=network_profile,
-                                                      storage_profile=storage_profile))
-
+    try:
+      compute_client.virtual_machines.create_or_update(
+        resource_group, vm_network_name, VirtualMachine(location=zone,
+                                                        os_profile=os_profile,
+                                                        hardware_profile=hardware_profile,
+                                                        network_profile=network_profile,
+                                                        storage_profile=storage_profile))
+    except CloudError as error:
+      raise AgentRuntimeException("There was a problem creating the "
+                                  "Virtual Machine due to the error: {0}"
+                                  .format(error.message))
     # Sleep until an IP address gets associated with the VM.
     while True:
       public_ip_address = network_client.public_ip_addresses.get(resource_group,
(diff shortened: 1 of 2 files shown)


appscale/tools/agents/azure_agent.py
```diff
@@ -265,31 +266,77 @@ def describe_instances(self, parameters, pending=False):
     private_ips = []
     instance_ids = []
 
-    public_ip_addresses = network_client.public_ip_addresses.list(resource_group)
-    for public_ip in public_ip_addresses:
-      public_ips.append(public_ip.ip_address)
+    public_ip_addresses, error = self.run_and_return_exception(
```
tmarballi: Is this the case whether we could possibly run into an exception while trying to just list the IP addresses in the resource group? The one you said @whoarethebritons you ran into while testing? And this is basically us being defensive right or did we always run into it?

appscale/tools/agents/azure_agent.py
```diff
@@ -265,31 +266,77 @@ def describe_instances(self, parameters, pending=False):
...
+        public_ip_addresses, raise_exception=True,
+        err_msg="Error calling describe instances")
+
+    public_ips = [x.ip_address for x in public_ips_results]
+
+    network_interfaces, error = self.run_and_return_exception(
+        network_client.network_interfaces.list, resource_group)
+    self.raise_if_error(error, "Error calling describe instances")
+
+    network_interfaces_results = self.convert_to_list(
```
tmarballi: I like the convert_to_list :), its especially nice for Azure because we have to go through a bunch of loops to get some information at times.

appscale/tools/agents/azure_agent.py
whoarethebritons: @tmarballi yes exactly I ran into it in testing, one of the reasons is because describe instances is now a part of run_instances in the Infrastructure Manager so we need to be able to handle a "too many requests" issue.

appscale/tools/agents/azure_agent.py
whoarethebritons: 😄

cdonati:
@whoarethebritons Sorry, but I am not a fan of the `run_and_return_exception`, `raise_if_error` pattern you are using here. It's essentially equivalent to a bare except, which will catch even more than `except Exception` (see https://www.python.org/dev/peps/pep-0008/#programming-recommendations, "When catching exceptions...").

I also don't see a reason to avoid using try/except.

whoarethebritons:
@cdonati ok, I'm going to close this PR then and make a new branch for it.

PR #675, opened 2018-02-12
Allow different instance type per role

tmarballi:
Needs https://github.com/AppScale/appscale/pull/2672


tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/4662/

appscale/tools/remote_helper.py
```diff
@@ -288,40 +328,32 @@ def spawn_load_balancers_in_cloud(cls, options, agent, params, count=1):
...
-    This method also prepares the virtual machine for use by the AppScale Tools.
+    try:
+      instance_ids, public_ips, private_ips = agent.run_instances(
+        count=count, parameters=params, security_configured=True,
+        public_ip_needed=load_balancer)
+    except (AgentRuntimeException, BotoServerError):
+      AppScaleLogger.warn("AppScale was unable to start the requested number "
+                          "of instances, attempting to terminate those that "
+                          "were started.")
+      if len(spawned_instance_ids) > 0:
```
cdonati: Can the caller do this?

appscale/tools/remote_helper.py
```diff
@@ -151,31 +153,68 @@ def start_all_nodes(cls, options, node_layout):
...
-      options, agent, params,
-      len(load_balancer_nodes))
+    load_balancer_roles = {}
 
-    for node_index, node in enumerate(load_balancer_nodes):
-      index = node_layout.nodes.index(node)
-      node_layout.nodes[index].public_ip = public_ips[node_index]
-      node_layout.nodes[index].private_ip = private_ips[node_index]
-      node_layout.nodes[index].instance_id = instance_ids[node_index]
+    instance_type_roles = {'with_disks':{}, 'without_disks': {}}
```
cdonati: I'm having trouble figuring out the reason for this separation.

whoarethebritons: @cdonati azure cannot attach existing disks to scaleset machines so they would have to be created as regular machines, it was added in a little early with the expectation of managed disks.

tmarballi:
@jeanleonov I did test the autoscaling on Azure, euca and GCE. I can record those for a demo or do them live if you prefer that. Let me know if there's any specific test that you would like to see to confirm it works as it should.

appscale/tools/remote_helper.py
cdonati: Okay. That sounds like an implementation detail that belongs in the azure agent instead of here.

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/4736/

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/4745/

cdonati:
@tmarballi Did you see my question about the caller performing the terminate? I think it would be unexpected for a spawn function to terminate other instances as a side effect.

tmarballi:
@cdonati Oops, I completely missed that. Looking into it now.

appscale/tools/remote_helper.py
```diff
@@ -151,27 +153,72 @@ def start_all_nodes(cls, options, node_layout):
...
+      instance_type.setdefault(node.instance_type, []).append(node)
+
+    spawned_instance_ids = []
+
+    for instance_type, load_balancer_nodes in load_balancer_roles.items():
+      # Copy parameters so we can modify the instance type.
+      instance_type_params = params.copy()
+      instance_type_params['instance_type'] = instance_type
+
+      instance_ids, public_ips, private_ips = cls.spawn_nodes_in_cloud(
```
cdonati: Does this need to clean up after itself if it encounters an exception?

tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/4758/

appscale/tools/remote_helper.py
tmarballi: Yes it does, It left my mind that we could technically have different instance types for LBs and it would need clean up if the 2nd LB faced a problem after the 1st one was started successfully. I have added the same try except as for non load balancer nodes.

appscale/tools/remote_helper.py
```diff
@@ -850,6 +886,28 @@ def wait_for_machines_to_finish_loading(cls, host, keyname):
...
+
+    Args:
+      spawned_instance_ids: A list of instance ids we have started that
+        should be terminated.
+      agent: The agent to call terminate instance with.
+      params: Agent parameters.
+    """
+    terminate_params = params.copy()
+    terminate_params[agent.PARAM_INSTANCE_IDS] = spawned_instance_ids
+    for _ in range(len(spawned_instance_ids)):
```
cdonati: Why is this running the terminate call multiple times (based on the number of started machines)?

tmarballi: You are right, it does not need a loop because the terminate instances method uses a list of instances to delete anyways.

tmarballi:
Latest all blue build: https://ocd.appscale.com:8080/job/Daily%20Build/4767/

PR #686, opened 2018-04-18
Consolidate configuration parsing functionality

### LINKED ISSUE #683: Consolidate configuration parsing functionality
Currently, we re-find and re-parse the application configuration files for every element we need. It would be nice if we had a single class to represent a configuration resource.

### LINKED ISSUE #685: Provide more informative error when config is missing
Currently, if the user's app does not have an app.yaml or appengine-web.xml, they will get the following error message during a deploy: "coercing to Unicode: need string or buffer, NoneType found".

### LINKED ISSUE #690: Support "service" keyword in app.yaml
At the moment, we just support the "module" keyword.

cdonati:
Currently, we re-find and re-parse the application configuration files for every element we need. The duplicated error handling and file reading code was difficult to maintain.

This adds a single class to handle parsing of app.yaml and appengine-web.xml files. As a bonus, this does not require provided archives to be unpacked to disk.

Resolves #683 
Resolves #685 
Resolves #690


appscale/tools/appscale_tools.py
```diff
@@ -959,60 +955,50 @@ def upload_app(cls, options):
...
+      if version.configuration_type == 'app.yaml':
+        message = 'Specify --project or define "application" in your app.yaml'
+      else:
+        message = 'Define "application" in your appengine-web.xml'
 
-        # Java App Engine users may have specified their war directory. In that
-        # case, just move up one level, back to the app's directory.
-        file_location = file_location + os.sep + ".."
-        app_id = AppEngineHelper.get_app_id_from_app_config(file_location)
+      raise AppEngineConfigException(message)
```
jeanleonov: What's difference between `AppEngineConfigException` and `BadConfigurationException`?

appscale/tools/admin_api/version.py
```diff
@@ -0,0 +1,327 @@
+""" Represents an Admin API Version resource. """
+
+from __future__ import absolute_import
```
jeanleonov: I still can't understand why we get ImportError on absolute imports of appscale.tools modules unless we import this future feature.

appscale/tools/appscale_tools.py
cdonati: I think originally `AppEngineConfigException` was just for the app.yaml and appengine-web.xml while `BadConfigurationException` could be for other things like the AppScalefile. It doesn't look like we really stuck to that, though. I think they should be consolidated.

appscale/tools/admin_api/version.py
cdonati: I don't think it's needed in this particular file. When I started working on this branch, this file started out in appscale.tools.

This import has the side effect of disabling the old relative import syntax. In Python 2, you can import a different module in the current module with either `import other_module` or `import .other_module` (newer relative import syntax). Since the namespace package is named "appscale" and there is an "appscale" module in appscale.tools, using `import appscale` from a module in "appscale.tools" is ambiguous. It could either refer to the other module on the same level as the current one or the root "appscale" module.

jeanleonov: Ohh.. appscale module!
Thanks now I understand=)

appscale/tools/appscale_tools.py
```diff
@@ -15,35 +16,30 @@
...
-import yaml
-from xml.etree import ElementTree
-
 from collections import Counter
 from itertools import chain
+from xml.etree import ElementTree
 
-# AppScale-specific imports
-from tabulate import tabulate
+import yaml
```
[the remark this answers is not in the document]
cdonati: I just used PyCharm's "Optimize Imports" feature. PyYAML is not in the standard library, so it puts it in a different section.

appscale/tools/appscale_tools.py
cdonati: In some cases, it would fail on both. In others, it would fail just on AppScale. I would be favor of changing that exception name, or even just replacing everything with `BadConfiguration`, but I think it should go in a separate pull request.

PR #687, opened 2018-04-18
Don't suppress timeouts when making AC calls

### LINKED ISSUE #684: 'appscale status' returns 'is up' for a terminated cloud deployment
Noticed on an old deployment: doing `appscale status` (with or without `-v`) can return success:

> `AppScale is up. All 0 nodes are loaded`

Even though the deployment has been torn down. In this case the login node IP has been re-used and is ping-able, but SSH into it with creds in .appscale is not possible.

cdonati:
Resolves #684

### DIFF
--- a/appscale/tools/appcontroller_client.py
+++ b/appscale/tools/appcontroller_client.py
@@ -1,24 +1,18 @@
 #!/usr/bin/env python
 
+from __future__ import absolute_import
 
-# General-purpose Python library imports
 import json
-import socket
 import signal
+import socket
 import ssl
 import time
 
-
-# Third-party imports
 import SOAPpy
 
-
-# AppScale-specific imports
-from appscale_logger import AppScaleLogger
-from custom_exceptions import AppControllerException
-from custom_exceptions import AppScaleException
-from custom_exceptions import BadSecretException
-from custom_exceptions import TimeoutException
+from appscale.tools.appscale_logger import AppScaleLogger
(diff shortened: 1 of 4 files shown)


appscale/tools/appcontroller_client.py
```diff
@@ -461,32 +551,43 @@ def create_user(self, username, password, account_type='xmpp_user'):
...
-          'Request to create user timed out.', self.DEFAULT_NUM_RETRIES,
-          self.server.create_user, username, password, account_type, self.secret)
+        result = self.run_with_timeout(
+          self.LONGER_TIMEOUT, self.DEFAULT_NUM_RETRIES,
+          self.server.create_user, username, password, account_type,
+          self.secret)
         break
-      except Exception, exception:
+      except Exception as exception:
         AppScaleLogger.log("Exception when creating user: {0}".format(exception))
```
sjones4: Should this catch the timeout exception and break the loop?

sjones4: I think my concern was mainly due to the possibility of unbounded retries rather than the handling of timeouts specifically.

cdonati: I agree that we probably don't want infinite retries. I'll open another PR after this one is merged to give those loops a limit.

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5148/

Issue #688, opened 2018-04-24
Fix SecretStorage dependency version


jeanleonov:
`msrestazure` requires `keyring` which requires `SecretStorage`. SecretStorage stopped maintaining Python 2.7 starting from its 3.0.0 version. To fix it we have explicitly require SecretStorage<3.

PR #689, opened 2018-04-24
Explicitely require SecretStorage<3 to use python 2.7

jeanleonov:
https://github.com/AppScale/appscale-tools/issues/688
https://ocd.appscale.com:8080/job/Daily%20Build/4949/
Rebuild migration - https://ocd.appscale.com:8080/job/Migration/2803/


jeanleonov:
Thanks for review, but 3 days ago [msrestazure](https://github.com/Azure/msrestazure-for-python) fixed this problem on their side by removing `keyring` dependency which was using `SecretStorage`. So we don't need this dependency at all.

PR #706, opened 2018-08-09
Azure error handling

whoarethebritons:
Attempts to fix a bug where exceptions thrown in threads are unhandled and provides more reliability with our infrastructure manager (for now, pending further changes to it).


appscale/tools/agents/azure_agent.py
```diff
@@ -717,78 +793,156 @@ def terminate_instances(self, parameters):
...
-          vmss_delete_threads.append(thread)
-
-      for delete_thread in vmss_delete_threads:
-        delete_thread.join()
+      vmss_delete_futures = []
+      vmss_delete_exceptions = set()
+
+      try:
+        ss_to_delete = [vmss.name for vmss in vmss_list if
+                       sum(1 for _ in
```
cdonati:
```
[vmss.name for vmss in vmss_list
 if not list(compute_client.virtual_machine_scale_set_vms.list(resource_group, vmss.name))]
```
is probably a bit simpler to read.

appscale/tools/agents/azure_agent.py
```diff
@@ -829,16 +978,40 @@ def delete_vmss_instance(self, compute_client, parameters, vmss_name, instance_i
...
     resource_group = parameters[self.PARAM_RESOURCE_GROUP]
     verbose = parameters[self.PARAM_VERBOSE]
-    AppScaleLogger.verbose("Deleting Virtual Machine Instance {0} from Scale "
-      "Set {1} ...".format(instance_id, vmss_name), verbose)
+
+    vm_info = "Virtual Machine {0} from Scale Set {1}".format(instance_id,
+                                                              vmss_name)
+
+    # Check if we succeeded deleting the instance before but timed out.
+    vm_list = compute_client.virtual_machine_scale_set_vms.list(
```
cdonati: Requiring the client to check first if the instance exists before issuing the delete call seems cumbersome. It's also possible for the instance to be created after the list and before the delete.

What happens if you issue the delete for an instance that doesn't exist?

appscale/tools/agents/azure_agent.py
```diff
@@ -829,16 +978,40 @@ def delete_vmss_instance(self, compute_client, parameters, vmss_name, instance_i
...
     result = compute_client.virtual_machine_scale_set_vms.delete(resource_group,
                                                                  vmss_name,
                                                                  instance_id)
     resource_name = 'Virtual Machine Instance ' + instance_id
     self.sleep_until_delete_operation_done(result, resource_name,
                                            self.MAX_VM_UPDATE_TIME, verbose)
-    AppScaleLogger.verbose("Virtual Machine Instance {0} from Scale Set {1} "
-      "has been successfully deleted".format(instance_id, vmss_name), verbose)
+    # Double check if we succeeded deleting the instance.
+    vm_list = compute_client.virtual_machine_scale_set_vms.list(
```
cdonati: Is this really necessary? Is it a known problem for the `delete` call to succeed and the instance to survive?

whoarethebritons: I have seen it print out that we successfully deleted an instance when the instance is still there.

appscale/tools/agents/azure_agent.py
whoarethebritons: We waste about 6 minutes waiting for the instance to be deleted, if I remember correctly.

whoarethebritons: I can double check on this though.

appscale/tools/agents/azure_agent.py
scragraham: s/list/any/
`if not any(..) ` might perform better, doesn't have to create a list.

appscale/tools/agents/azure_agent.py
whoarethebritons: there are no consequences to calling the delete without listing, so I will change this.

whoarethebritons:
https://ocd.appscale.com:8080/job/Daily%20Build/5381/

PR #714, opened 2018-09-14
Create load balancer VMs in an availability set

tmarballi:
This created nodes with load balancer roles within an availability set. So that we can place an Azure load balancer which uses the availability set to distribute traffic between multiple load balancers machines.


tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/5410/

appscale/tools/agents/azure_agent.py
```diff
@@ -446,13 +453,34 @@ def run_instances(self, count, parameters, security_configured, public_ip_needed
...
+
+    # Create an Availability Set for Load Balancer VMs
+    lb_avail_set_name = "lb-availability-set"
+    availability_set_names = [availability_set.name for availability_set in
+                              compute_client.availability_sets.list(resource_group)]
+    if lb_avail_set_name not in availability_set_names:
+        lb_avail_set = self.create_lb_availability_set(compute_client,
+                                                       lb_avail_set_name,
+                                                       parameters)
+        if lb_avail_set:
```
cdonati: It looks to me like `self.create_lb_availability_set` raises an exception on error rather than return `None`.

appscale/tools/agents/azure_agent.py
```diff
@@ -446,13 +453,34 @@ def run_instances(self, count, parameters, security_configured, public_ip_needed
...
+    compute_client = ComputeManagementClient(credentials, subscription_id)
+
     subnet = self.create_virtual_network(network_client, parameters,
                                          virtual_network, virtual_network)
 
     active_public_ips, active_private_ips, active_instances = \
       self.describe_instances(parameters)
+
+    # Create an Availability Set for Load Balancer VMs
+    lb_avail_set_name = "lb-availability-set"
```
cdonati: Wouldn't we want want different deployments to use different availability sets?

appscale/tools/agents/azure_agent.py
```diff
@@ -446,13 +453,34 @@ def run_instances(self, count, parameters, security_configured, public_ip_needed
...
+    availability_set_names = [availability_set.name for availability_set in
+                              compute_client.availability_sets.list(resource_group)]
+    if lb_avail_set_name not in availability_set_names:
```
cdonati: `availability_set` needs to be defined if this condition isn't met since it's used below.

appscale/tools/agents/azure_agent.py
whoarethebritons: ^ and also on a successful update it would still return an `AvailabilitySet`

appscale/tools/agents/azure_agent.py
```diff
@@ -486,8 +515,36 @@ def run_instances(self, count, parameters, security_configured, public_ip_needed
...
+        compute_client: A ComputeManagementClient instance
+        lb_avail_set_name: The name to use for the availability set.
+        parameters: A dict, containing all the parameters necessary to
+          authenticate this user with Azure.
+    Raises:
+        AgentConfigurationException: If the operation to create an availability
+          set did not succeed.
+    """
+    try:
+      return compute_client.availability_sets.create_or_update(
```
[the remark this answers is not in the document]
tmarballi: Yeah I too didn't know why only this operation does not follow the usual returning AzureOperationPoller type behavior.

appscale/tools/agents/azure_agent.py
tmarballi: Good point. I have fixed it in the commit https://github.com/AppScale/appscale-tools/pull/714/commits/09709d4d3ef96e33ce7b40756d04b3d5c02250e0. It doesn't appear to resolve the changes here before the line I changed is different than where you pointed at.

appscale/tools/agents/azure_agent.py
```diff
@@ -486,8 +515,36 @@ def run_instances(self, count, parameters, security_configured, public_ip_needed
...
+        parameters[self.PARAM_RESOURCE_GROUP], lb_avail_set_name,
+        AvailabilitySet(location=parameters[self.PARAM_ZONE]))
+    except CloudError as error:
+      logging.exception("Azure agent received a CloudError while creating an "
+                        "availability set.")
+      raise AgentConfigurationException("Unable to create an Availability Set "
```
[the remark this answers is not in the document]
tmarballi: Yeah, I will change that.

appscale/tools/agents/azure_agent.py
tmarballi: So when we create an availability set, it is basically similar to having a folder for all the load balancer VMs to be placed within and then those VMs are removed from it when the deployment is terminated. I was of the opinion that the same availability set can just be reused between different deployments, because their only use right now is to have a set of load balancer VMs which then can be used by the Azure load balancer to distribute load to.

I can use a lib to automatically generate a new name instead of hard coding it here but I thought this name would be more relevant because we would only have one per deployment.

cdonati: So if you have two deployments running at the same time and they are both using the same availability set, how does the load balancer know which VMs should get which traffic?

tmarballi: Oh so this is per resource group in Azure. And we can currently only do one deployment per resource group in Azure. I meant the same availability set name to be used between deployments started and terminated within the same resource group. Outside of the resource group, even if there exists another same named availability set within another resource group it wouldn't matter, because all the deployment resources are bound with the resource group under within they are created. So there would not be any clashes.

cdonati: Thanks, that makes sense.

tmarballi:
Latest build: https://ocd.appscale.com:8080/job/Daily%20Build/5469/

PR #716, opened 2018-09-20
Include scaling information from app.yaml or appengine-web.xml for version create

sjones4:
Application versions created from YAML or XML now include manual or automatic scaling information when present. The client sends the scaling information in the version create call for use when scaling the application.

https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.services.versions#Version


appscale/tools/admin_api/version.py
```diff
@@ -76,6 +78,22 @@ def from_yaml(app_yaml):
...
     version.inbound_services = app_yaml.get('inbound_services', [])
 
+    automatic_scaling = app_yaml.get('automatic_scaling', None)
+    manual_scaling = app_yaml.get('manual_scaling', None)
+    if manual_scaling:
+      try:
+        version.manual_scaling = {'instances': int(manual_scaling['instances'])}
+      except StandardError:
+        raise AppEngineConfigException('Invalid app.yaml: manual_scaling invalid.')
+    elif automatic_scaling:
```
cdonati: It seems this has an unexpected behavior if the user mistakenly defines both.

I would vote to either include all defined parameters in the request (and let the Admin API reject it) or reject it here.

This is minor and shouldn't block, though.

sjones4:
Added a check for exclusive manual/automatic scaling settings.

PR #717, opened 2018-09-26
Add VPC network support for AWS

whoarethebritons:
with classic networking:
https://ocd.appscale.com:8080/job/EC2-Testing/3
with vpc networking:
https://ocd.appscale.com:8080/job/EC2-Testing/6

requires AppScale/appscale#2877


appscale/tools/agents/ec2_agent.py
```diff
@@ -147,26 +156,33 @@ def configure_instance_security(self, parameters):
...
-    security_groups = conn.get_all_security_groups()
-    for security_group in security_groups:
-      if security_group.name == group:
-        self.handle_failure("Security group {0} is already registered. Please" \
-          " change the 'group' specified in your AppScalefile to a different " \
-          "value, or erase it to have one automatically generated for you." \
-          .format(group))
+    try:
+      self.get_security_group_by_name(parameters, group)
+      self.handle_failure("Security group {0} is already registered. Please "
```
cdonati: This should be in an `else` clause after the except. Otherwise, the `AgentRuntimeException` that `handle_failure` raises will be caught by the except.

From a readability standpoint, it also makes it clearer that the call is supposed to fail (a success being a more exceptional event).

appscale/tools/agents/ec2_agent.py
```diff
@@ -477,6 +582,13 @@ def run_instances(self, count, parameters, security_configured, public_ip_needed
         AppScaleLogger.log("Waiting for your instances to start...")
         public_ips, private_ips, instance_ids = self.describe_instances(
           parameters)
+
+        # If we need a public ip, make sure we actually get one.
+        if public_ip_needed and not self.diff(public_ips, private_ips):
+          time.sleep(self.SLEEP_TIME)
+          now = datetime.datetime.now()
```
cdonati: To make this simpler and prevent future updates here from accidentally forgetting to update `now`, just replace `now` in the if statement and remove the assignments.

appscale/tools/agents/ec2_agent.py
```diff
@@ -796,12 +908,19 @@ def cleanup_state(self, parameters):
 
     AppScaleLogger.log("Deleting security group {0}".format(
       parameters[self.PARAM_GROUP]))
-    while True:
+    retries_left = self.SECURITY_GROUP_RETRY_COUNT
+    while retries_left:
```
cdonati: If an EC2ResponseError is raised more than the number of retries, this function will return successfully.

appscale/tools/agents/ec2_agent.py
```diff
@@ -200,37 +220,69 @@ def create_security_group(self, parameters, group):
...
+      group: A str that names the group that should be found. Or None to
+        use the group in parameters.
+    Returns:
+      The 'boto.ec2.securitygroup.SecurityGroup' that has the correct group
+      name.
+    Raises:
+      AgentRuntimeException: If the security group could not be found.
+    """
+    # Get security group.
+    if not group:
```
cdonati: Given the name of the function, I don't think client should be permitted to omit the group name.

appscale/tools/agents/ec2_agent.py
whoarethebritons: would:
```
        if retries_left == self.SECURITY_GROUP_RETRY_COUNT:
          raise
```

in the except clause be sufficient you think?

cdonati: That should be fine. However, since you're counting down, you'll probably want to compare to 0 rather than the retry count. You'll also likely want to change to `while True` since you are moving your exit condition.

appscale/tools/agents/ec2_agent.py
```diff
@@ -796,12 +905,22 @@ def cleanup_state(self, parameters):
...
+        conn.delete_security_group(group_id=sg.id)
         return
       except EC2ResponseError:
-        time.sleep(5)
+        time.sleep(self.SLEEP_TIME)
+        retries_left -= 1
+      except AgentRuntimeException:
+        AppScaleLogger.log('Could not find security group {}, delete '
+                           'successful.'.format(parameters[self.PARAM_GROUP]))
+        if retries_left == 0:
```
cdonati: Shouldn't this be in the other except clause?

appscale/tools/agents/ec2_agent.py
whoarethebritons: Can I mark this one as resolved?

appscale/tools/agents/ec2_agent.py
whoarethebritons: Just want to make sure I'm understanding correctly, you mean do `while datetime.datetime.now() < end_time:`?

cdonati: Yes, that's right.

appscale/tools/agents/ec2_agent.py
whoarethebritons:
```suggestion
  def get_security_group_by_name(self, parameters, group):
    """Gets a security group in AWS with the given name.
     Args:
      parameters: A dict that contains the credentials necessary to authenticate
        with AWS.
      group: A str that names the group that should be found.
    Returns:
      The 'boto.ec2.securitygroup.SecurityGroup' that has the correct group
      name.
    Raises:
      EC2ResponseError: If there was an error contacting EC2.
      AgentRuntimeException: If the security group could not be found.
    """
    conn = self.open_connection(parameters)
    try:
      return next(sg for sg in conn.get_all_security_groups()
                  if sg.name == group)
    except StopIteration:
      raise AgentRuntimeException('Could not find security group with name '
                                  '{}!'.format(group))

```

and then in 
```suggestion
      except EC2ResponseError:
        retries_left -= 1
        if retries_left == 0:
          break
        time.sleep(self.SLEEP_TIME)
        
```
and leave the AgentRuntimeException one to return if it couldn't find it?

and also catch the EC2ResponseError in other methods that call get_security_group_by_name

cdonati: Looks like we edited our comments at roughly the same time :smile: 

I think what you proposed is a step in the right direction. You probably want to `raise` instead of `break` when the retries are up, though. I also still think `AgentRuntimeException` shouldn't be used to indicate that the SG does not exist since it is used in other places to indicate a failure or transient error.

appscale/tools/agents/ec2_agent.py
```diff
@@ -191,43 +213,70 @@ def create_security_group(self, parameters, group):
       "name {0}".format(group))
 
 
-  def authorize_security_group(self, parameters, group, from_port, to_port,
-    ip_protocol, cidr_ip):
+  def get_security_group_by_name(self, conn, group):
+    """Gets a security group in AWS with the given name.
```
sjones4: Security group names are unique for classic networking or for a given vpc. To find a group by name you would need to check that the vpc matched (or that the group was not a vpc group)

appscale/tools/agents/ec2_agent.py
```diff
@@ -191,43 +213,70 @@ def create_security_group(self, parameters, group):
...
         on.
       to_port: An int that names the last port that access should be allowed on.
       ip_protocol: A str that indicates if TCP, UDP, or ICMP traffic should be
         allowed.
       cidr_ip: A str that names the IP range that traffic should be allowed
         from.
+      group_name: A str that contains the name of the group whose ports
+        should be opened. Default is None since EC2 can just use group_id but
+        Euca will override to the group name in the parameters. EC2 cannot
+        use both group_id and group_name.
```
sjones4: It seems simpler to use group_id in all cases.

whoarethebritons: group_id didn't appear to work for me in euca for classic networking, so ec2 is using group_id always but euca calls the method using the group name.

appscale/tools/agents/ec2_agent.py
whoarethebritons: So I need to double check that the vpc ids match?

sjones4: My understanding is that we do not currently check the vpc for the security group during look up and only check that the group name matches.

If looking for a group in a vpc you would check the vpc id of the group matched the configured vpc for appscale. If looking for a classic networking security group (i.e. no configured vpc) you would check that the group was not in a vpc (had no vpc id)

appscale/tools/agents/ec2_agent.py
sjones4: I think the issue is that the euca agent is configured to use the `2010-08-31` ec2 api which was different from the one being used in the current `boto` client. You could probably just bump the euca agent to use the current api version '2016-11-15'.

whoarethebritons:
with vpc:
https://ocd.appscale.com:8080/job/EC2-Testing/26/
without vpc:
https://ocd.appscale.com:8080/job/EC2-Testing/27/

https://ocd.appscale.com:8080/job/Daily%20Build/5596/
https://ocd.appscale.com:8080/job/Datastore/269/
https://ocd.appscale.com:8080/job/Migration/3269/

PR #721, opened 2018-10-16
Use managed availability set for adding VMs created with managed disks

tmarballi:
The type of availability set currently in master does not work with VMs created with managed disks. So in this PR, if the VMs are created using the marketplace public image then a managed availability set gets created and if not the regular one which works with VHD images.


tmarballi:
https://ocd.appscale.com:8080/job/Daily%20Build/5486/

tmarballi:
From what I know of managed disks, the only way to currently use them in our Azure agent is to use it with a marketplace image. Managed disks can only be used with a publisher image and not with our custom AppScale vhd image. And @whoarethebritons correct me if I am wrong, but that's the only way we can attach disks currently for any Azure deployment right?

PR #724, opened 2018-10-17
Appscale services start/stop commands (serving status modification)

sjones4:
Adds start and stop command for application/project/service (default version):

```
appscale services start <project_id> <service_id>
appscale services stop <project_id> <service_id>
```

This modifies the serving status for the application version.

Relates to Appscale/appscale#2894

### DIFF
--- a/appscale/tools/admin_api/client.py
+++ b/appscale/tools/admin_api/client.py
@@ -210,34 +210,20 @@ class AdminClient(object):
     return operation_id
 
   @retry(**RETRY_POLICY)
-  def delete_project(self, project_id):
-    """ Deletes a project.
-
-    Args:
-      project_id: A string specifying the project ID.
-    Raises:
-      AdminError if the response is not 200.
-    """
-    url = 'https://{}:{}/v1/projects/{}'.format(self.host, self.PORT,
-                                                project_id)
-    headers = {'AppScale-Secret': self.secret}
-    response = requests.delete(url, headers=headers, verify=False)
-    if response.status_code != 200:
-      raise AdminError('Error asking Admin Server to delete project!')
-
-  @retry(**RETRY_POLICY)
-  def list_projects(self):
-    """ Lists projects.
+  def list_services(self, project_id):
+    """ Lists a project's services.
 
     Returns:
-      A list containing the projects of this deployment.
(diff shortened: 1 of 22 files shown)


appscale/tools/admin_api/client.py
```diff
@@ -147,6 +147,45 @@ def delete_version(self, project_id, service_id, version_id):
...
+                    .format(prefix=self.prefix, project=version.project_id,
+                            service=version.service_id, version=version.id))
+    headers = {
+      'AppScale-Secret': self.secret,
+      'Content-Type': 'application/json'
+    }
+    parameters = {
+      'updateMask': ','.join(fields)
+    }
+    query_parameters = '&'.join('='.join(item) for item in parameters.items())
```
cdonati: I would recommend using `urllib.urlencode(parameters)` for this use case. An alternative would be using the `params` argument to `requests.patch`.

appscale/tools/appscale_tools.py
```diff
@@ -630,6 +631,65 @@ def remove_app(cls, options):
...
+        passed in via the command-line interface.
+    Raises:
+      AppScaleException: If the named service isn't running in this
+        AppScale cloud, or if start is not valid for the service.
+    """
+    login_host = LocalState.get_login_host(options.keyname)
+    secret = LocalState.get_secret_key(options.keyname)
+    admin_client = AdminClient(login_host, secret)
+
+    version = Version(None, None)
```
cdonati: Should this just be `Version(None)`? It looks like it just takes the runtime parameter.

sjones4: I think `Version(None, None)` is correct for `master`:

```
class Version(object):
  """ Represents an Admin API Version resource. """
  def __init__(self, runtime, config_type):
```

https://github.com/AppScale/appscale-tools/blob/master/appscale/tools/admin_api/version.py#L22

cdonati: You're right. I was looking at an older commit.

appscale/tools/admin_api/client.py
sjones4: Updated to pass parameters directly to `requests.patch`. Thanks for the suggestion.

PR #725, opened 2018-10-17
EC2 infrastructure agent disk device / warning

sjones4:
Update the ec2 agent to allow for `/dev/vd*` devices and the euca agent to not require the `IS_VERBOSE` parameter. This allows use of `infrastructure : euca` with persistent disks.

The warning for not using persistent disks is no longer shown when disks are configured under `ips_layout`.

### DIFF
--- a/appscale/tools/agents/euca_agent.py
+++ b/appscale/tools/agents/euca_agent.py
@@ -54,7 +54,7 @@ class EucalyptusAgent(EC2Agent):
       self.handle_failure('Unknown scheme in EC2_URL: ' + result.scheme)
       return None
 
-    if parameters.get('IS_VERBOSE', None):
+    if parameters.get(self.PARAM_VERBOSE, False):
       debug_level = 2  # extremely verbose
     else:
       debug_level = 0  # the silent treatment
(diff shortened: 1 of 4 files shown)


sjones4:
https://ocd.appscale.com:8080/job/Daily%20Build/5496/

appscale/tools/agents/euca_agent.py
```diff
@@ -54,7 +54,7 @@ def open_connection(self, parameters):
       self.handle_failure('Unknown scheme in EC2_URL: ' + result.scheme)
       return None
 
-    if parameters['IS_VERBOSE']:
+    if parameters.get('IS_VERBOSE', None):
```
cdonati: Very minor Python style nitpick, but `None` is the default value anyways for `.get`, so `parameters.get('IS_VERBOSE')` works here.

`parameters.get('IS_VERBOSE', False)` would also be fine since an explicit boolean is a bit more readable in the context of an "if" statement.

appscale/tools/node_layout.py
```diff
@@ -513,6 +513,19 @@ def db_master(self):
...
     return None
 
+  def are_disks_used(self):
+    """ Searches through the nodes in this NodeLayout to see if any persistent
+    disks are being used.
+
+    Returns:
+      True if any persistent disks are used, and False otherwise.
+    """
+    disks = [node.disk for node in self.nodes]
```
cdonati: This is perfectly acceptable, but I'd just like to share the convenience of `any`:
`return any(node.disk for node in self.nodes)`

PR #726, opened 2018-10-24
Update cryptography dependency to a recent version

sjones4:
Bump the `cryptography` dependency to  2.3.0 minimum, avoiding known security issues with some intermediate versions. This requires updating `pyOpenSSL` to v18 which no longer supports python 2.6.


sjones4:
https://ocd.appscale.com:8080/job/Daily%20Build/5540/

PR #728, opened 2018-10-25
Remove a project's services on 'undeploy'

### LINKED ISSUE #727: Undeploying a project should not delete it
Since we don't have a mechanism in place to delete all of a particular project's data and metadata, it would make more sense for 'undeploy' to delete a project's services rather than all of a project's metadata.

cdonati:
Since the project delete operation is no longer supported, delete the project's services instead.

Resolves #727.

### DIFF
--- a/appscale/tools/admin_api/client.py
+++ b/appscale/tools/admin_api/client.py
@@ -174,34 +174,20 @@ class AdminClient(object):
     return operation_id
 
   @retry(**RETRY_POLICY)
-  def delete_project(self, project_id):
-    """ Deletes a project.
-
-    Args:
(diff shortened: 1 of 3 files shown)


sjones4:
We may want to avoid using the term 'stop' for removed/deleted services and just use 'delete'. Stop may be confused with stop/start for app versions.

https://cloud.google.com/sdk/gcloud/reference/app/services/delete
https://cloud.google.com/sdk/gcloud/reference/app/versions/stop

cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5632/
https://ocd.appscale.com:8080/job/Migration/

PR #734, opened 2018-12-09
Fix index configuration parsing for Java apps

cdonati:
Previously, the wrong element was being selected as the XML root.

This also adds the SDK-generated "datastore-indexes-auto.xml" as a fallback source for index configuration.


cdonati:
https://ocd.appscale.com:8080/job/Daily%20Build/5731/

PR #738, opened 2019-01-28
Have get_params_from_args set defaults if an arg is not defined

whoarethebritons:
https://ocd.appscale.com:8080/job/Create-Azure-Image/466/console
https://ocd.appscale.com:8080/job/Hawkeye-Test-Azure-Image/183/console

https://ocd.appscale.com:8080/job/Daily%20Build/5861/

### DIFF
--- a/appscale/tools/agents/azure_agent.py
+++ b/appscale/tools/agents/azure_agent.py
@@ -95,6 +95,9 @@ class AzureAgent(BaseAgent):
   # Default resource group name to use for Azure.
   DEFAULT_RESOURCE_GROUP = 'appscalegroup'
 
+  # Default resource tag name to use for Azure.
+  DEFAULT_TAG = 'default-tag'
+
   # A list of Azure instance types that have less than 4 GB of RAM, the amount
   # recommended by Cassandra. AppScale will still run on these instance types,
   # but is likely to crash after a day or two of use (as Cassandra will attempt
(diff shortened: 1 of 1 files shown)


appscale/tools/agents/azure_agent.py
```diff
@@ -1473,29 +1476,25 @@ def get_params_from_args(self, args):
       self.PARAM_INSTANCE_TYPE: args[self.PARAM_INSTANCE_TYPE],
       self.PARAM_GROUP: args[self.PARAM_GROUP],
       self.PARAM_KEYNAME: args[self.PARAM_KEYNAME],
-      self.PARAM_RESOURCE_GROUP: args[self.PARAM_RESOURCE_GROUP],
-      self.PARAM_STORAGE_ACCOUNT: args[self.PARAM_STORAGE_ACCOUNT],
+      self.PARAM_RESOURCE_GROUP: args.get(self.PARAM_RESOURCE_GROUP,
+                                          self.DEFAULT_RESOURCE_GROUP),
```
cdonati: What is the advantage of having a default resource group as opposed to failing if the resource group is not defined? I would imagine most users do not have access to 'appscalegroup'.

whoarethebritons: I would be fine with failing without a resource group, I was just moving the logic we use below for assigning a default.

scragraham:
Just so I don't forget: This also needs to be merged to appscale/appscale-agents

PR #741, opened 2019-02-05
Whitelist additional acceptable instance types

cdonati:
This re-adds first-generation instances as well as a few newer-generation instances.


cdonati:
Demo (hawkeye single with m1.large)
With master: https://ocd.appscale.com:8080/job/Hawkeye/17240/
With branch: https://ocd.appscale.com:8080/job/Hawkeye/17235/

PR #743, opened 2019-02-25
appscale.tools.agents replacement with appscale.agents

scragraham:
- Removal of agents code in appscale.tools
 - Added some additional Makefile recipes.


Makefile
```diff
@@ -1,13 +1,51 @@
 #
 # Makefile is mainly used to perform a version bump on the software.
 #
-.PHONY: bump-major bump-minor bump-patch help all
+.PHONY: bump-major bump-minor bump-patch help all test
+
+AGENTS_REPO?=https://github.com/scragraham/appscale-agents
+AGENTS_BRANCH?=topic-agent-init
```
sjones4: are these defaults correct?

scragraham: no they are not, thanks. fixed.

PR #746, opened 2019-03-09
Upgrade cloud fix

whoarethebritons:
With a multi-node cloud deployment an upgrade will fail since it will use "node-0", "node-1", etc as the public ips.

The instance_type option is added from the AppScalefile since nodes are determined to be equal if they have the same roles and instance types.

### DIFF
--- a/appscale/tools/appscale.py
+++ b/appscale/tools/appscale.py
@@ -925,4 +925,5 @@ Available commands:
     options.ips = yaml.safe_load(base64.b64decode(options.ips_layout))
     options.terminate = False
     options.clean = False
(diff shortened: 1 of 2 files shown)


whoarethebritons:
Upgrade-Test with AppScale/appscale-tools#746
cloud: https://ocd.appscale.com:8080/view/All/job/Upgrade-Test/49/
local: https://ocd.appscale.com:8080/view/All/job/Upgrade-Test/50/

Upgrade-Test with master
cloud: https://ocd.appscale.com:8080/view/All/job/Upgrade-Test/51/
local: https://ocd.appscale.com:8080/view/All/job/Upgrade-Test/52/

PR #748, opened 2019-03-27
Require a keyring version with python 2 support

sjones4:
Keyring v19 dropped support for Python 2.7 - 3.4, an explicit dependency is added to ensure a compatible version is installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
     'google-api-python-client==1.5.4',
     'haikunator',
     'httplib2',
+    'keyring<19.0.0',
     'msrestazure==0.4.34',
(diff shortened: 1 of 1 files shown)


sjones4:
Build is 6064.

Demo:
Expected version is installed:
```
# pip list | grep keyring
keyring (18.0.1)
```
Error no longer logged when using `appscale` command:
```
# appscale init
AppScalefile successfully created! Be sure to customize it for your particular cloud or cluster.
```
Prior to this fix the output was:
```
# appscale init
No handlers could be found for logger "msrestazure.azure_active_directory"
AppScalefile successfully created! Be sure to customize it for your particular cloud or cluster.
```

PR #754, opened 2019-07-19
AppController client handle not ready error

sjones4:
Not sure how to reproduce this, but error handling could be improved by handling this error response rather than trying to parse this error as json (etc)


sjones4:
Daily build no. 6503

PR #759, opened 2019-08-07
Remove use of monit for appscale-controller, updates for systemd support

sjones4:
Use `systemctl` to start the controller.

This pull request should be merged along with AppScale/appscale#3135

Daily build no. 6600


sjones4:
Will open a new pull request when the full changes for monit removal are ready.

PR #760, opened 2019-08-09
Dispatch routing from yaml

whoarethebritons:
`appscale deploy` has been modified to accept a dispatch.yaml file and requires project to be set. The dispatch file is then read and the tools send dispatchRules in the format described [here](https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps#UrlDispatchRule) to the AdminServer but verification is done on the AdminServer side.

Known Issues:
- The AdminServer will respond instantly saying the dispatchRules were applied although it might take until the next duty cycle for them to be applied.

Build:
http://ocd.appscale.net:8080/job/Daily%20Build/6607/

Requires: AppScale/appscale#3136


appscale/tools/appscale_tools.py
```diff
@@ -1117,6 +1117,57 @@ def upload_app(cls, options):
...
+        continue
+
+      if 'error' in operation:
+        raise AppScaleException(operation['error']['message'])
+      dispatch_rules = operation['response']['dispatchRules']
+      break
+
+    AppScaleLogger.verbose(
+        "The following dispatchRules have been applied to your application's "
+        "configuration (note: rules have been converted to be compatible "
```
cdonati: Hopefully we aren't doing any conversion to the rules themselves, so this notice shouldn't be needed.

appscale/tools/utils.py
```diff
@@ -272,6 +276,24 @@ def queues_from_xml(contents):
 
   return queues
 
+def dispatch_from_yaml(source_location):
+  dispatch_rules = None
+  with open(source_location) as config_file:
+    dispatch_rules = yaml.safe_load(config_file.read())
```
cdonati: Super minor, but just as a tip, `yaml.safe_load` accepts the file descriptor object as an argument so you don't need to read it all in first.

PR #762, opened 2019-08-22
Remove azure / keyring dependencies

sjones4:
These dependencies are transitive via appscale-agents so can be removed here.

Relates to AppScale/appscale-agents#15


sjones4:
Daily build no. 6665

PR #763, opened 2019-08-29
Add --update flag with `appscale up` to specify code directory

tmarballi:
This allows to specify a code directory to update and build during an appscale up.


tmarballi:
Build: http://ocd.appscale.net:8080/job/Daily%20Build/6700/

appscale/tools/scripts/appscale.py
```diff
@@ -41,8 +41,26 @@ def main():
...
   elif command == "up":
+    update_dir = ""
+    if len(sys.argv) > 2:
+      if sys.argv[2] != '--update':
+        cprint("Usage: appscale up [--update] <code directory to update>", 'red')
+        sys.exit(1)
+
+      if len(sys.argv) > 4:
+        dir_args = sys.argv[3:]
+        update_dir = ",".join(dir_args)
```
cdonati: `update_dir` can be kept as a list

appscale/tools/scripts/appscale.py
```diff
@@ -41,8 +41,26 @@ def main():
...
+
+      else:
```
cdonati: A single argument does not need to be a special case here. You could simplify with something like
```
if len(sys.argv) < 4:
  (print usage)

update_dir = sys.argv[3:]
```

appscale/tools/parse_args.py
```diff
@@ -137,6 +137,10 @@ class ParseArgs(object):
   # explicitly provide a value, in megabytes.
   DEFAULT_MAX_APPSERVER_MEMORY = 400
 
+  # The list of code directories to specify which updating the code and building it.
+  ALLOWED_DIR_UPDATES = ["all", "common", "app_controller", "admin_server",
```
cdonati: This can be a tuple since nothing is modifying it.

appscale/tools/parse_args.py
```diff
@@ -183,6 +187,10 @@ def add_allowed_flags(self, function):
     #TODO: remove arguments in appscale-run-instances that are no longer
     # supported. (min, max, appengine, max_memory, scp)
     if function == "appscale-run-instances":
+      self.parser.add_argument('--update',
+        nargs='+',
+        choices=self.ALLOWED_DIR_UPDATES,
+        default="", help="updates specified code directory and builds it")
```
cdonati: I would recommend an empty list as a default so that you can be sure the parsed arg object is a list.

appscale/tools/local_state.py
```diff
@@ -217,6 +217,12 @@ def generate_deployment_params(cls, options, node_layout, additional_creds):
     }
     creds.update(additional_creds)
 
+    if options.update:
+      update_dir_creds = {
+        'update': str(options.update)
```
cdonati: Since this is all being serialized to JSON later, you can just use the list itself as for the "update" key.

appscale/tools/appscale.py
```diff
@@ -239,10 +239,12 @@ def init(self, environment=None):
     shutil.copy(self.TEMPLATE_APPSCALEFILE, appscalefile_location)
 
 
-  def up(self):
+  def up(self, update=""):
```
cdonati: I would recommend an empty tuple as a default arg here in order to match the expected iterable argument.

appscale/tools/appscale.py
```diff
@@ -239,10 +239,12 @@ def init(self, environment=None):
...
     """ Starts an AppScale deployment with the configuration options from the
     AppScalefile in the current directory.
 
+    Args:
+        update: An appscale code directory to update and build.
```
cdonati: A list of...

appscale/tools/appscale.py
```diff
@@ -257,6 +259,12 @@ def up(self):
 
     # Construct a run-instances command from the file's contents
     command = []
+    if update:
+        command.append("--update")
+        update_dirs = update.split(',')
+        for dir in update_dirs:
```
cdonati: `dir` is a keyword in Python, so it's usually best to avoid using it as a variable name.

To simplify things a bit, you could just do `command.extend(update)` (assuming `update` is an iterable).

appscale/tools/appscale.py
```diff
@@ -239,10 +239,12 @@ def init(self, environment=None):
     shutil.copy(self.TEMPLATE_APPSCALEFILE, appscalefile_location)
 
 
-  def up(self):
+  def up(self, update=[]):
```
cdonati: Default arguments shouldn't be mutable since any change to it will affect the default arg for future calls.

tmarballi:
Latest build: http://ocd.appscale.net:8080/job/Daily%20Build/6708/

PR #767, opened 2019-09-27
Remote logging, send only general deployment information when enabled

sjones4:
Add environment setting to disable remote logging; when enabled send only id, version, and state information and infrastructure and scaling options.


sjones4:
Daily build no. 6850

